PR #7, opened 2016-03-15
Pull request hip

sunway513:
Hello Simon,
This pull request is on behalf of HIP team from AMD, to add GPU-STREAM for HIP-compatible devices.
 
The first [commit](https://github.com/UoB-HPC/GPU-STREAM/commit/fdeb20601fc33d76383b2f833cb902f80225d025) contains the minimum support for running GPU-STREAM benchmark on HIP.
The second [commit](https://github.com/UoB-HPC/GPU-STREAM/commit/11053798ff7dca2bd8eb7a8e5bacac10758b5159) includes the improvement we have made to use optional looper kernel and take command-line parameters for groups and groupSize. We observed significant performance gain by using the looper kernels for single precision type. 

Please kindly review the pull request and let us know any questions or suggestions.

Thanks,
Peng


tomdeakin:
It would be great if you could clarify some things about HIP for us. From what we understand, HIP allows for a translation of CUDA source into HIP source, which you then compile and can link into a variety of back ends. As such, is the `hip-stream.cpp` file auto generated?

Also, it would be great if the addition of the results were in separate commits to the HIP source commit.

sunway513:
Hello Tom,
Thanks for the reply.

CUDA source code could be partially translated by the tool called hipify, more details have been introduced in the following blog:
http://gpuopen.com/hip-to-be-squared-an-introductory-hip-tutorial/

The hip-stream.cpp in the first commit was mostly auto-generated from the cuda-stream.cu file by hipify; while the looper optimizations was manually added in hip-stream.cpp in the second commit.

And sure, I will remove the results and submit them in a separate commit.
Please let us know if anything more we need to clarify regarding HIP and the pull request. 

Regards,
Peng

tomdeakin:
Hello Peng,

Thanks for clarifying a few things for us.

It would be good to see the looper kernels implemented in both the OpenCL and CUDA versions so that we can have a fair comparison, especially as you said this helps on some devices with single precision. It would be best to put this in two commits, one for each version.

Then, given the that HIP version is mostly auto generated with a new manual edits, is it possible that you could write a sed script for the generic HIP manual edits to the generated version, and let the Makefile completely auto generate the HIP code?

Thanks for separating the results commit from the source changes, but there is still a mix of results in the one commit. We would really like each result file to be it's own commit so we can pull in the results one-by-one.

bensander:
Hip is designed as a porting toolkit rather than a makefile step.   There are several reasons for this including :
- hip implements a well defined subset of Cuda.  Programmers who code to hip are guaranteed their program will run on both Nvidia and Amd palforms, but not all cuda programs can be directly / automatically translated to hip.
- hip has some features specifically designed to enabled porting.  A good example is the arch feature flags (i.e. hasDoubles) - the hip implementation is portable to either arch while Cuda offers no such guarantee.
- hip also uses wider 64bit int types for some cross lane instructions (i.e. Ballot, shuffle) to account for the option of a wider warp width such as used in AMD Fiji.

Overall we are seeing same performance for hip ports as compared to native cuda - the hip layer is very thin.  For projects that want a single maintainable code base we recommend a one time port to hip, then perform subsequent development in the hip repository. Many benchmarks have multiple back ends (i.e. Cuda, Opencl, openmp,  openacc) and for these cases it can be interesting to add hip to the mix for comparison purposes on performance, loc complexity, etc.

-Ben

On Mar 31, 2016, at 9:35 AM, Tom Deakin <notifications@github.com<mailto:notifications@github.com>> wrote:


Hello Peng,

Thanks for clarifying a few things for us.

It would be good to see the looper kernels implemented in both the OpenCL and CUDA versions so that we can have a fair comparison, especially as you said this helps on some devices with single precision. It would be best to put this in two commits, one for each version.

Then, given the that HIP version is mostly auto generated with a new manual edits, is it possible that you could write a sed script for the generic HIP manual edits to the generated version, and let the Makefile completely auto generate the HIP code?

Thanks for separating the results commit from the source changes, but there is still a mix of results in the one commit. We would really like each result file to be it's own commit so we can pull in the results one-by-one.

-

sunway513:
Hello Tom,
Thanks for your comments.
I have spitted the submitted results one per commits here for your convenience merging to the main repository. 
Besides, the new [commit](207701219a6346bf7f833f9f463fc9f93a6c3971) includes the looper optimization for cuda version. 
Please review the pull request and let me know if you have further comments.

Regards,
Peng

sunway513:
Hello Tom,
Did you have a chance to review our pull request recently?
Please let me know if there is anything else we want to discuss to get it merged.

Regards,
Peng

tomdeakin:
We're a bit confused about how  `CLUMP_SIZE` is set. It looks like it's only in the copy kernel, and isn't used in any of the others, and it's hard coded to 1 anyway. Are you able to describe in words what this parameter does?

We are about to embark on a significant rewrite of GPU-STREAM, which will be a major release. We will include a HIP version in this rewrite too, so we will merge in your commit after clarifying `CLUMP_SIZE` for us.

Thanks for your patience.

sunway513:
Hello Tom,
Thanks for the reply. We are glad to know you will have a major release soon on GPU-STREAM and HIP will be included in the rewrite.

Regarding your query, we reconsidered the CLUMP_SIZE usage in this scenario. We decide to remove the CLUMP_SIZE from the pull request for now, and will update this part when we have further support later.

Regards,
Peng

PR #12, opened 2016-10-18
Updated the SYCL Stream benchmark with latest ComputeCpp CE 0.1.1 Edition

Ruyk:
This patch updates the CMake building options to support the [Codelay ComputeCpp Community Edition 0.1](https://www.codeplay.com/products/computesuite/computecpp), including the FindComputeCpp CMake module provided with the [ComputeCpp SDK](https://github.com/codeplaysoftware/computecpp-sdk).
The patch also includes minor performance tuning changes for SYCL. First, we pre-compile the kernel binaries in the setting up, like the OCLStream variant. By default on SYCL, the OpenCL program is compiled on demand in the Parallel For. 
The accessors syntax is also changed to use linear access, which simplifies the work of the device compiler.
    
In order to build with ComputeCpp, only the standard CMake flags from the SDK are required:
 ` cmake ../
      -DHAS_SYCL=ON \
      -DCOMPUTECPP_PACKAGE_ROOT_DIR=/path/to/computecpp/package \
      -DCMAKE_MODULE_PATH=GPU-STREAM/cmake/Modules/`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,17 +155,22 @@ if (HAS_SYCL)
   add_executable(gpu-stream-sycl main.cpp SYCLStream.cpp)
   target_compile_definitions(gpu-stream-sycl PUBLIC SYCL)
 
-  find_program(COMPUTECPP "compute++")
-  if (COMPUTECPP)
+  # The user must define this in order to use FindComputeCpp
+  if (COMPUTECPP_PACKAGE_ROOT_DIR)
     message(STATUS "Using ComputeCpp for SYCL compilation")
-    add_custom_target(SYCLStream.sycl COMMAND ${COMPUTECPP} ${CMAKE_CURRENT_SOURCE_DIR}/SYCLStream.cpp -sycl -no-serial-memop -O2 -emit-llvm -c)
-    add_dependencies(gpu-stream-sycl SYCLStream.sycl)
-    target_compile_options(gpu-stream-sycl PUBLIC -include SYCLStream.sycl)
-    target_link_libraries(gpu-stream-sycl SYCL OpenCL)
+    include(FindComputeCpp)
+
+    include_directories(${COMPUTECPP_INCLUDE_DIRECTORY})
+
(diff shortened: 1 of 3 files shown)


jrprice:
The build system changes look fine, but I'm seeing a big performance drop with the changes to the SYCL kernels. On a Fury X we used to get ~430 GB/s, and with this patch we're only getting ~384 GB/s. We're using ComputeCpp CE0.1.

Any idea what's going on here?

Ruyk:
The code used explicit local size for SYCL, with a value that was good for AMD but quite bad for other platforms we are testing internally. The default behaviour works well for all platforms on the current development version of ComputeCpp. We are planning update the public version of ComputeCpp soon. We can wait to merge this patch until the update is public and you verify the results, or I can add the explicit values only for AMD devices in the patch, so that you can merge this patch in now, whatever you prefer :smiley:

jrprice:
Ah, I see. We like the idea of not having to specify the work-group size. If there's a new version of ComputeCpp being released soon that gives us a good performance with that approach, then we'll just hang on for a bit longer and merge this patch after it is released.

Ruyk:
ComputeCpp 0.1.1 is now out :-)

jrprice:
Thanks - performance is now just a little higher than before.

jrprice:
Actually I spoke too soon. The performance is slightly better on the AMD Fury X , but for Intel CPUs it's *much* worse. For our Ivy Bridge Xeon machine it's almost twice as slow, and for our Intel Xeon Phi (Knights Landing) box it's 15X slower (a mere 15 GB/s down from 214 GB/s).

I'm not sure why this would be. For the OpenCL version we just pass `NULL` to `clEnqueueNDRangeKernel` and let the runtime pick the work-group size, which works just fine. My understanding was that this new version of ComputeCpp should be doing the same?

jrprice:
False alarm, I just screwed something up. Performance on Intel devices is no worse than before.

Ruyk:
Thanks!

PR #13, opened 2016-11-02
Replace write_arrays with init_arrays

jrprice:
This allows each model to initialise their arrays with a parallel
approach, which yields the first touch required for good performance
on NUMA architectures.

With this change, I wonder whether we should now make it so that the model classes all "own" their own host arrays, rather than passing in the vectors from `main.cpp`. The driver would still need to allocate a vector to retrieve the results, but it seems a little strange to allocate and pass in a vector which quite a few models now just ignore.

### DIFF
--- a/ACCStream.cpp
+++ b/ACCStream.cpp
@@ -36,13 +36,19 @@ ACCStream<T>::~ACCStream()
 }
 
 template <class T>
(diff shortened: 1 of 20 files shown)


tomdeakin:
I agree with your comment about letting the models own their own host arrays. Really main should just read back the result at the end.

tomdeakin:
Merged into the devel branch

PR #19, opened 2016-12-09
Unify OpenMP 4.5 code for CPU and GPU

tomdeakin:
This update defaults to running OpenMP on the CPU. If built with `-DOMP_TARGET_GPU` the OpenMP 4.5 target code is built instead.

Additionally the `-DOMP3` and `-DOMP45` defines are replaced with just `-DOMP`.


OMPStream.h
```diff
@@ -17,7 +17,7 @@
 #define IMPLEMENTATION_STRING "OpenMP 4.5"
```
jrprice: Should we make this just `"OpenMP"` now?

Issue #18, opened 2016-11-26
enable OpenMP 4.5 for compilers besides Cray


jeffhammond:
I'm not a CMake guru but this project should test for OpenMP 4.5 compiler support generically rather than just associate this implementation with the Cray toolchain.

http://www.openmp.org/resources/openmp-compilers/ has the full list, but at least GCC 6.1+ and Intel 17+ support OpenMP 4.5 and I just confirmed that GCC 6.2.0 can correctly run `gpu-stream-omp45`.

```
$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index efee733..574fcc3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,12 +118,12 @@ endif ()
 #-------------------------------------------------------------------------------
 # OpenMP 4.5
 #-------------------------------------------------------------------------------
-if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Cray")
-  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.5)
+#if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Cray")
+#  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.5)
     add_executable(gpu-stream-omp45 main.cpp OMP45Stream.cpp)
     target_compile_definitions(gpu-stream-omp45 PUBLIC OMP45)
-  endif ()
-endif ()
+#  endif ()
+#endif ()
```

tomdeakin:
This should be solved in v3.1. Additionally in v3.0 we changed the structure on the OpenMP code to allow building for CPUs and GPUs via compiler preprocessor switch rather than separate source.

Issue #14, opened 2016-11-13
Build system needs improving


tomdeakin:
Binaries don't tend to build with the correct flags. This needs fixing. In the mean time things can be compiled manually, but the following should be run first:

`echo '#define VERSION_STRING "2.0"' > common.h`

tomdeakin:
Fixed in v3.1.

Issue #9, opened 2016-08-26
Make fails for RAJA


tomdeakin:
CUDA and RAJA interfeer in CMakeLists.txt

DavidPoliakoff:
Hey, I'm running into this as well. I get RAJA to build if I add

  set_source_files_properties(main.cpp PROPERTIES CUDA_SOURCE_PROPERTY_FORMAT OBJ)

But then the OMP3 build crashes. Don't know whether it's possible to fix along those lines.

tomdeakin:
We haven't found a way of getting RAJA to play nicely with CMake; we also find that CUDA breaks if you add that line. There are lots of other issues with the CMake too as lots of flags are missing for many of the models. We are just working out what it going to work best and what is needed across the range of devices across the platforms we use.

In the mean time, we're building manually with this command:
`nvcc --expt-extended-lambda -arch=sm_52 -O3 -std=c++11 -DUSE_RAJA main.cpp RAJAStream.cpp -x cu -L$RAJA_PATH/lib -lRAJA  -Xcompiler -fopenmp`

You may also have to make sure `common.h` exists. It just contains
`#define VERSION_STRING 2.1`

DavidPoliakoff:
Hey Tom, thanks for the update. 

In our current implementation we have an (everchanging) custom FindCUDA.cmake which we use to handle a lot of ugliness (including how to do CUDA compilation with the Clang CUDA compiler). I might file an issue on our repo about this, it seems like the kind of thing that ought to work.

Honestly, some of the problem is that we're all reconciling different CUDA oddities, and resolving them in different ways. Let me get back to this when I have time, but thanks again, you've brought a lot of good feedback to the project, it's highly appreciated

tomdeakin:
Fixed in v3.1. We've moved away from CMake so can better control how we build against RAJA. New version works for all combinations we tried in our usual results matrix.

Issue #8, opened 2016-07-25
GTX1080 CUDA issues


psteinb:
I wanted to benchmark a GTX 1080 with cuda 8.0.27 under CentOS 7.2.1511. the gpu-stream-cuda app behaves normal with the default parameters. 
Strange enough though, when I want to provide more than the default number of elements in the array:

```
$ gpu-stream-cuda --arraysize 67108864
```

the copy kernel dispatch throws a CUDA API error `0xb` which is `Invalid Argument`. I tracked down the problem to (this line of code)[https://github.com/UoB-HPC/GPU-STREAM/blob/master/CUDAStream.cu#L112]:

```
template <class T>
void CUDAStream<T>::copy()
{
  copy_kernel<<<array_size/TBSIZE, TBSIZE>>>(d_a, d_c);
  check_error();
  cudaDeviceSynchronize();
  check_error();
}
```

strange enough, if I look at the values of array_size/TBSIZE, they are in plausible ranges `arraysize/TBSIZE = 65536`.

Does anyone have an idea where this is coming from? (as this is a RC cuda, I see no problem forwarding this issue to nvidia)

psteinb:
happens with cuda 7.5 on a K80 as well

michaelboulton:
I think that the maximum for block sizes is 65535, not 65536. Might be the issue?

psteinb:
are you sure? `deviceQuery` from cuda 8 SDK reports something else:
```
Device 0: "GeForce GTX 1080"
  CUDA Driver Version / Runtime Version          8.0 / 8.0
  CUDA Capability Major/Minor version number:    6.1
  Total amount of global memory:                 8113 MBytes (8507555840 bytes)
  (20) Multiprocessors, (128) CUDA Cores/MP:     2560 CUDA Cores
  GPU Max Clock rate:                            1848 MHz (1.85 GHz)
  Memory Clock rate:                             5005 Mhz
  Memory Bus Width:                              256-bit
  L2 Cache Size:                                 2097152 bytes
  Maximum Texture Dimension Size (x,y,z)         1D=(131072), 2D=(131072, 65536), 3D=(16384, 16384, 16384)
  Maximum Layered 1D Texture Size, (num) layers  1D=(32768), 2048 layers
  Maximum Layered 2D Texture Size, (num) layers  2D=(32768, 32768), 2048 layers
  Total amount of constant memory:               65536 bytes
  Total amount of shared memory per block:       49152 bytes
[...]

PR #24, opened 2017-02-23
Simplify build system

tomdeakin:
This branch aims to simplify the build system in general for GPU-STREAM. CMake is unable to build multiple binaries using multiple compilers with a single evocation. As such, it is simpler to just provide simple Makefiles which build the binaries for the various models in common configurations. They are simple to extend and customise for specific platforms if required.

This should fix a number of the Issues, in particular #8, #9, #14 and #18.


SYCL.make
```diff
@@ -0,0 +1,11 @@
+
+sycl-stream: main.cpp SYCLStream.cpp SYCLStream.sycl
+	$(CXX) -O3 -std=c++11 -DSYCL main.cpp SYCLStream.cpp -include SYCLStream.sycl -lComputeCpp -lOpenCL -o $@
+
+
+SYCLStream.sycl: SYCLStream.cpp
+	compute++ SYCLStream.cpp -sycl -no-serial-memop -O2 -emit-llvm -c
```
Ruyk: If computecpp_info is on the path (as compute++ seems to be) you can just query it for the device compiler flags:
```
$ computecpp_info  --dump-device-compiler-flags
-O2 -mllvm -inline-threshold=1000 -sycl -intelspirmetadata -emit-llvm
```
which you can use then for the makefile

jrprice: Thanks, we'll use this instead.

This doesn't include `-no-serial-memop` though (which leads to warnings from the compiler). Should we just ignore that?

Issue #25, opened 2017-02-27
HIP dot product fails to build


psteinb:
Hi, I just wanted to start adding a bare HC stream and started off with the HIPified version in cc90cef and got this:

```
$ $ make -f HIP.make 
hipcc  -std=c++11 -DHIP main.cpp HIPStream.cu  -o hip-stream
HIPStream.cu:185:10: error: cannot combine with previous 'extern' declaration specifier
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
         ^
/opt/rocm/include/hip/hcc_detail/host_defines.h:49:24: note: expanded from macro '__shared__'
#define __shared__     tile_static
                       ^
/opt/rocm/hcc/include/hc_defines.h:52:21: note: expanded from macro 'tile_static'
#define tile_static static __attribute__((section("clamp_opencl_local")))
                    ^
HIPStream.cu:185:31: error: expected parameter declarator
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
                              ^
HIPStream.cu:185:31: error: expected ')'
HIPStream.cu:185:30: note: to match this '('
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
                             ^
HIPStream.cu:185:21: error: C++ requires a type specifier for all declarations
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
  ~~~~~~~~~~~~~~~~~ ^
HIPStream.cu:185:21: error:  tile_static can only be applied to a variable declaration
HIPStream.cu:185:41: error: expected ';' at end of declaration
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
                                        ^
                                        ;
HIPStream.cu:185:56: error: definition of variable with array type needs an explicit size or an initializer
  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
                                                       ^
HIPStream.cu:188:11: error: use of undeclared identifier 'blockDim'
  int i = blockDim.x * blockIdx.x + threadIdx.x;
          ^
HIPStream.cu:188:24: error: use of undeclared identifier 'blockIdx'
  int i = blockDim.x * blockIdx.x + threadIdx.x;
                       ^
HIPStream.cu:188:37: error: use of undeclared identifier 'threadIdx'
  int i = blockDim.x * blockIdx.x + threadIdx.x;
                                    ^
HIPStream.cu:189:26: error: use of undeclared identifier 'threadIdx'
[...]

PR #26, opened 2017-02-28
Fix sharedmem hip

psteinb:
fix of #25

### DIFF
--- a/HIPStream.cpp
+++ b/HIPStream.cpp
@@ -1,5 +1,3 @@
-
-
 // Copyright (c) 2015-16 Tom Deakin, Simon McIntosh-Smith,
 // University of Bristol HPC
 //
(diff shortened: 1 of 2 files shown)


HIPStream.cpp
```diff
@@ -1,3 +1,5 @@
+
+
```
jrprice: Spurious extra lines added?

HIPStream.cpp
```diff
@@ -177,22 +184,22 @@ void HIPStream<T>::triad()
...
-
 template <class T>
 __global__ void dot_kernel(hipLaunchParm lp, const T * a, const T * b, T * sum, unsigned int array_size)
 {
 
-  extern __shared__ __align__(sizeof(T)) unsigned char smem[];
-  T *tb_sum = reinterpret_cast<T*>(smem);
+  HIP_DYNAMIC_SHARED(T,tb_sum);
+  // HIP_DYNAMIC_SHARED(unsigned char,smem);
+  // T *tb_sum = reinterpret_cast<T*>(smem);
```
jrprice: Please remove this commented-out code if not needed.

jrprice:
Thanks!

jrprice:
FYI the dynamic shared memory approach here broke things for NVIDIA, so I've switched both CUDA and HIP to use static shared memory instead. I *think* the HIP version should still work fine with HCC now.

PR #27, opened 2017-02-28
replaced - for = so that assignment takes place

psteinb:
not that it changes anything in the results AFAIK, just for completeness


tomdeakin:
Good spot! Merging.

PR #28, opened 2017-03-13
added EXTRA_FLAGS variable to CUDA Makefile to provide the freedom to…

psteinb:
… specify debug flags or gencode flags, 

in more detail, I ran into trouble when trying to benchmark more than `arraysize = 32*1024*1024`. This is due to the fact that by default, not `-gencode` flags or similar are setup for the nvcc call of the cuda GPU stream. This defaults to an SM that doesn't support an arraysize of this magnitude and hence the `init_arrays` fails with error code `0xb Invalid arguments`. We had the discussion about gencodes before in #8, but the new Makefile based build system doesn't reflect the outcome of #8.

### DIFF
--- a/CUDA.make
+++ b/CUDA.make
@@ -1,6 +1,7 @@
+EXTRA_FLAGS?=-O3
 
 cuda-stream: main.cpp CUDAStream.cu
-	nvcc -std=c++11 -O3 -DCUDA $^ $(EXTRA_FLAGS) -o $@
+	nvcc -std=c++11 -DCUDA $^ $(EXTRA_FLAGS) -o $@
 
 .PHONY: clean
 clean:
--- a/CUDA.make
+++ b/CUDA.make
@@ -1,7 +1,7 @@
-EXTRA_FLAGS?=-O3
+CXXFLAGS?=-O3 -std=c++11
 
 cuda-stream: main.cpp CUDAStream.cu
-	nvcc -std=c++11 -DCUDA $^ $(EXTRA_FLAGS) -o $@
+	nvcc $(CXXFLAGS) -DCUDA $^ $(EXTRA_FLAGS) -o $@
 
 .PHONY: clean
 clean:
--- a/CUDA.make
+++ b/CUDA.make
@@ -1,7 +1,8 @@
-CXXFLAGS?=-O3 -std=c++11
+CXXFLAGS=-O3
+CUDA_CXX=nvcc
 
 cuda-stream: main.cpp CUDAStream.cu
-	nvcc $(CXXFLAGS) -DCUDA $^ $(EXTRA_FLAGS) -o $@
+	$(CUDA_CXX) -std=c++11 $(CXXFLAGS) -DCUDA $^ $(EXTRA_FLAGS) -o $@
 
 .PHONY: clean
 clean:


tomdeakin:
You don't need to predefine `EXTRA_FLAGS` in the Makefile for this to work. You can just specify the `gencode` at the command line with no changes:
```
make -f CUDA.make EXTRA_FLAGS="-g -gencode=<...>`
```

psteinb:
might be, but that wouldn't have switched off `-O3` that was hard coded 
in the Makefile stub.

tomdeakin:
That's true. Are you able to put the `-O3` flag in `CXXFLAGS` instead of `EXTRA_FLAGS` to be consistent with the other Makefiles, e.g `OpenMP.make`?

psteinb:
While I am at it, I'll align the cuda make to OpenMP.cmake although I'd 
recommend using `?=` where possible. OpenMP.cmake does always set 
variables directly where `?=` might offer more flexiblity.

I'll finish the CUDA.make and thus leave the door open for CUDA-with-clang.

tomdeakin:
What flexibility are you seeing with `?=`? If you override any variables on the CLI then Make still uses this user defined one and ignores all the assignments with `=` in the Makefile.

psteinb:
With this:

```
CXXFLAGS?=-O3 -std=c++11
GPUCXX?=nvcc

cuda-stream: main.cpp CUDAStream.cu
         $(GPUCXX) $(CXXFLAGS) -DCUDA $^ $(EXTRA_FLAGS) -o $@
```

I can pitch in anything on the command line and the Makefile stays quite 
concise:

```
$ CXXFLAGS="-g -G -std=c++11" make -f CUDA.cmake #produces debug output 
in case I screwed up
$ make -f CUDA.cmake #produces the default -O3 binary
```

Just saying -

jrprice:
Sure, but the same is true without the `?`. e.g. if the Makefile looks like this:

```
CXXFLAGS=-std=c++11 -O3

cuda-stream: main.cpp CUDAStream.cu
        nvcc ${CXXFLAGS} -DCUDA $^ $(EXTRA_FLAGS) -o $@
```

You can still override `CXXFLAGS` like this:

    make -f CUDA.cmake CXXFLAGS="-g -G -std=c++11"

psteinb:
NB. I still need to dig the GNU make manual.

Ha, learned something again! I checked your example with a handful of 
GNU make builds:

# with `?=`

```
$ cat testme.make
CXXFLAGS?=-std=c++11 -O3

cuda-stream: main.cpp CUDAStream.cu
		nvcc ${CXXFLAGS} -DCUDA $^ $(EXTRA_FLAGS) -o $@
```

using `?=` is agnostic of the position of the variable definition:

```
$ CXXFLAGS=-bla make -f testme.make -n
nvcc -bla -DCUDA main.cpp CUDAStream.cu  -o cuda-stream
$ make -f testme.make -n CXXFLAGS=-bla
nvcc -bla -DCUDA main.cpp CUDAStream.cu  -o cuda-stream
```

## without `?=`

```
$ cat testme.make
CXXFLAGS=-std=c++11 -O3

cuda-stream: main.cpp CUDAStream.cu
		nvcc ${CXXFLAGS} -DCUDA $^ $(EXTRA_FLAGS) -o $@
```

Now, watch this:

~~~
$  make -f testme.make -n CXXFLAGS=-bla
nvcc -bla -DCUDA main.cpp CUDAStream.cu  -o cuda-stream
$ CXXFLAGS=-bla make -f testme.make -n
nvcc -std=c++11 -O3 -DCUDA main.cpp CUDAStream.cu  -o cuda-stream
~~~

See the last call? the updated value of CXXFLAGS is not used.

psteinb:
we didn't come to a conclusion on this, did we?

tomdeakin:
Thanks for your investigation into `=` vs `?=` - we'd not realised the implication of this before. I think we will want to move all the Makefiles to using `?=` instead of `=` so that it doesn't matter what position in the command line the variables are.

But for the purposes of this pull request, can you pull out the `-O3` flag to `CXXFLAGS` but use an `=` for now. We can go though and update all the Makefiles consistently to `?=` later.

psteinb:
done

tomdeakin:
Merged, thanks for the contribution.

tomdeakin:
Created Issue #30 for switching to `?=`

PR #31, opened 2017-03-24
added OpenCL for the ROCm platform

psteinb:
```
GPU-STREAM
Version: 3.1
Implementation: OpenCL
Running kernels 100 times
Precision: double
Using OpenCL device gfx803
Driver: 1.1 (HSA,LC)
Reduction kernel config: 256 groups of size 256
Copy        409471.915  0.00131     0.00137     0.00133     
Mul         409624.689  0.00131     0.00137     0.00134     
Add         418105.745  0.00193     0.00219     0.00200     
Triad       417470.030  0.00193     0.00228     0.00199     
Dot         461460.829  0.00116     0.00126     0.00120
```
created a specific target for OpenCL on the rocm platform. The only difference to vanilla gcc are the CXXFLAGS. But consider this as a placeholder for future rocm releases when hcc can be used to link OpenCL binaries as well.

### DIFF
--- a/OpenCL.make
+++ b/OpenCL.make
@@ -3,22 +3,23 @@ ifndef COMPILER
 define compiler_help
 Set COMPILER to change flags (defaulting to GNU).
 Available compilers are:
-  GNU CLANG INTEL CRAY
+  GNU CLANG INTEL CRAY HCC
 
 endef
 $(info $(compiler_help))
 COMPILER=GNU
 endif
 
+COMPILER_HCC = g++
 COMPILER_GNU = g++
 COMPILER_CLANG = clang++
 COMPILER_INTEL = icpc
 COMPILER_CRAY = CC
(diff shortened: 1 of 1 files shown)


Ruyk:
Which hardware is gfx803 ?

psteinb:
that is a R9 Fiji Nano with 4GB of HBM1.

jrprice:
I'm not sure I understand the need for this change. The host compiler being used is still GNU not HCC, so the only actual difference is the addition of a non-standard include path. This seems like it could just be achieved with `make -f OpenCL.make COMPILER=GNU EXTRA_FLAGS="-I/opt/rocm/opencl/include"`, rather than adding a specific configuration for it.

psteinb:
true, but this way it's more integrated. feel free to merge or close the PR if you think this is not needed.

PR #32, opened 2017-04-03
Rocm hc support

psteinb:
added support for GPU stream implemented with HC on the ROCm platform with AMD GPUs.

```
$ ./hc-stream --device 1
GPU-STREAM
Version: 3.1
Implementation: HC
Running kernels 100 times
Precision: double
Using HC device AMD HSA Agent gfx8032
Copy        408840.162  0.00131     0.00135     0.00133     
Mul         413124.761  0.00130     0.00133     0.00131     
Add         418120.941  0.00193     0.00196     0.00194     
Triad       419383.219  0.00192     0.00207     0.00194     
Dot         317894.801  0.00169     0.00183     0.00173
```

with a AMD R9 Fiji Nano on ubuntu 14.04.5 with rocm 1.4


psteinb:
should I update this PR given the recent tag of v3.2?

tomdeakin:
I don't think we've changed anything that affects this HC version.

psteinb:
very good, would you consider merging it then?

HC.make
```diff
@@ -0,0 +1,14 @@
+
+# TODO: HC with HCC
```
jrprice: What does this TODO reference? Isn't this Makefile already using HCC?

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  // The array size must be divisible by TBSIZE for kernel launches
+  if (ARRAY_SIZE % TBSIZE != 0)
+  {
+    std::stringstream ss;
+    ss << "Array size must be a multiple of " << TBSIZE;
+    throw std::runtime_error(ss.str());
+  }
+
+  // // Set device
```
jrprice: Nit: extra `//`

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+  std::vector<hc::accelerator> accs = hc::accelerator::get_all();
+  auto current = accs[device_index];
```
jrprice: Could use a sanity check here that `device_index` is in range.

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+                                });
+
+  hc::completion_future future_b= hc::parallel_for_each(hc::extent<1>(array_size)
+                                                        , [=](hc::index<1> i) [[hc]] {
+                                                          view_b[i] = _b;
+                                                        });
+  hc::completion_future future_c= hc::parallel_for_each(hc::extent<1>(array_size)
+                                                        , [=](hc::index<1> i) [[hc]] {
+                                                          view_c[i] = _c;
+                                                        });
```
jrprice: Is there a reason why these initialisation kernels are separate rather than just being a single lambda?

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  try{
+    hc::completion_future future_kernel = hc::parallel_for_each(hc::extent<1>(array_size)
+                                , [=](hc::index<1> i) [[hc]] {
+                                  view_b[i] = scalar*view_c[i];
+								});
+    future_kernel.wait();
+  }
+  catch(std::exception& e){
+    std::cerr << __FILE__ << ":" << __LINE__ << "\t HCStream<T>::copy " << e.what() << std::endl;
```
jrprice: `s/copy/mul/`

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  try{
+    hc::completion_future future_kernel = hc::parallel_for_each(hc::extent<1>(array_size)
+                                , [=](hc::index<1> i) [[hc]] {
+                                  view_c[i] = view_a[i]+view_b[i];
+								});
+    future_kernel.wait();
+  }
+  catch(std::exception& e){
+    std::cerr << __FILE__ << ":" << __LINE__ << "\t HCStream<T>::copy " << e.what() << std::endl;
```
jrprice: `s/copy/add`

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  try{
+    hc::completion_future future_kernel = hc::parallel_for_each(hc::extent<1>(array_size)
+                                , [=](hc::index<1> i) [[hc]] {
+                                  view_a[i] = view_b[i] + scalar*view_c[i];
+								});
+    future_kernel.wait();
+  }
+  catch(std::exception& e){
+    std::cerr << __FILE__ << ":" << __LINE__ << "\t HCStream<T>::copy " << e.what() << std::endl;
```
jrprice: `s/copy/triad`

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+template <class T>
+T HCStream<T>::dot()
+{
+   //implementation adapted from
+    //https://ampbook.codeplex.com/SourceControl/latest
+    // ->Samples/CaseStudies/Reduction
+    // ->CascadingReduction.h
+
+    static constexpr std::size_t n_tiles = 64;
```
jrprice: Where does this number come from? Is this something that could/should be determined dynamically from the device specification?

HCStream.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        for (auto h = TBSIZE / 2; h; h /= 2) {
+            if (tidx.local[0] < h) {
+                tileData[tidx.local[0]] += tileData[tidx.local[0] + h];
+            }
+            tidx.barrier.wait_with_tile_static_memory_fence();
+        }
+
+        if (tidx.global == tidx.tile_origin) partial[tidx.tile] = tileData[0];
+    });
```
jrprice: Is this implementation the same as the OpenCL/CUDA versions? It looks like it might be, but I'm not familiar enough with HC terminology to be sure.

HC.make
psteinb: fixed with b1fc309f3ac9953ddf3e8f93c2ae60ddb7ae9c66

HCStream.cpp
psteinb: fixed with f6d6874

HCStream.cpp
psteinb: fixed with a971591, would you like to have a try-else clause as well? or is the `accs.at(device_index)` enough?

HCStream.cpp
psteinb: no, not really ... would you prefer a merged initialisation? (this way is more easy to debug as there is not device side debugger available in rocm 1.4)

HCStream.cpp
psteinb: fixed in c0559c2

HCStream.cpp
psteinb: fixed in c0559c2

HCStream.cpp
psteinb: fixed in c0559c2

HCStream.cpp
psteinb: @AlexVlx Can you comment on the 64 (this is your code)? is it related to the wavefront size?

HCStream.cpp
psteinb: AFAIK, yes

HCStream.cpp
AlexVlx: Hello Peter. This is not quite my code, if you recall you picked it up from the C++AMP book, and I merely provided an alternative that worked in your context. It should indeed be dynamically determined, based on how many elements one may want to process per lane e.g.:
    - n_tiles = domain_size / (lanes_per_tile * elements_per_lane);
Note that the current value (64) is sub-optimal for modern AMD GPUs, such as e.g. Fiji. Thank you.

psteinb:
Fair enough, that slipped my memory. I apologize.

So then the number of elements per lane is the configurable parameter 
here. In BabelStream's CUDA implementation, the thread block size is 
fixed to 1024 by means of a preprocessor definition as is the grid size 
to 256, i.e. the reduction is grid size in nature. I am not sure what 
the reasoning was behind that as it also results in a dynamic number of 
work items per thread per device (as the number of active threads 
depends on the number of SMs etc) depending on what card you run on 
which can lead to quite some unfair comparisons if I am not mistaken.

tomdeakin:
Hard coding 1024 for CUDA comes from firstly having to choose a number, and therefore secondly  [some work](https://www.karlrupp.net/2016/01/gpu-memory-bandwidth-vs-thread-blocks-cuda-workgroups-opencl/) which says 1024 is a reasonable number. It's perhaps difficult to come up with a robust heuristic using only CUDA API calls to work out what this number should be in a device agnostic manner. OpenCL gets around this by not requiring a local size being set and letting the vendor choose in the runtime.

For the OpenCL reductions, we use the API calls to make a reasonable heuristic/guess which works well on all devices we've tested. We *could* use a similar approach for CUDA reductions, however it means we cannot use a static sized shared memory allocation. A change here would also need to be reflected in the HIP code.
I'm not convinced the evidence shows it makes a difference for the reduction. For the `dot` kernel, the CUDA and OpenCL results are *very similar* on NVIDIA GPUs, despite each code taking a different approach for determining the number of thread blocks/work groups to launch.

psteinb:
ok, so does that imply, that for hc we can stick to a *good number* that 
is able to reproduce e.g. the HIP or OpenCL results? As for `TBSIZE` 
this can be made configurable through preprocessor definitions. How 
about that?

On the topic of `TBSIZE` itself, the link you shared, shows that 
choosing 1024 makes the results go into compute bound part of the 
roofline model (above/equals 16 thread blocks), which is where (given 
the simplicity of c[:] = b[:] + a[:]) you are able to measure the memory 
bandwidth. this can only happen once the device is "fully" occupied in 
terms of memory loads/stores per SM. Which (for me) raises the concern, 
if a naive user runs the CUDA Babel stream with all default values, it 
maybe that he/she ends up in the bandwidth bound region of the roofline 
plot ... which clearly would not reflect the maximum bandwidth available 
from the card.

psteinb:
just did a small study on this with a GTX1080 and cuda 8.0.61:
cuda_dot_sweep
the default array size of 33554432 puts you in the compute bound section for this card, does that hold for all cards? (my apologies if I am nit picking here)

Ruyk:
Just to note that, if a change is made on the OpenCL code to alter array sizes or selection of local group sizes, the same custom selection of sizes should be done for the SYCL variant of the code. I am happy to help to make that change if required.
However, seems that having this "golden values" for each variant will favour vendor-specific approaches (such as CUDA or ROCm, that target specific hardware thus this golden values are easily determined) vs more generic approaches (OpenCL, SYCL, others...) where a golden value targeting one platform may hinder performance in others.

tomdeakin:
@Ruyk I'm happy with the current OpenCL and SYCL approach which (for reductions) queries the API and for GPU launches `4*CL_DEVICE_MAX_COMPUTE_UNITS` work-groups of size `CL_DEVICE_MAX_WORK_GROUP_SIZE`, and for CPU launches `CL_DEVICE_MAX_COMPUTE_UNITS` work-groups of size `2* CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE`. Essentially we're launching a "status quo" number of work-groups per physical hardware "thing" - 4 groups per compute unit on a GPU and 1 group per compute unit (core) on a CPU.

For non-reduction kernels it's a non-issue in OpenCL and SYCL as we can use a local size of `NULL`.

For both kernel types in CUDA we **have** to pick a "golden value" under the CUDA model. 1024 threads per thread block is such a number that could be classed as "best practice", which is what the focus of this code is: "If we follow best practices, how do we do?".


@psteinb STREAM is a memory bandwidth bound kernel, so I'm not sure I understand how you categorise it based on problem size, particularly with your reference to Roofline. The graph you plot is not Roofline as the axis are wrong. Roofline is FLOPS (y-axis) vs AI (x-axis), and the AI of the STREAM kernels are fixed no matter what problem size you run.

The rules of original STREAM say the arrays have to be larger than last level cache. At one point we set the rules of BabelStream similarly (possibly in an old README but I can't find it right now). However on GPUs this is an awkward metric, and so when benchmarking one device in isolation you should increase the problem size until it changes little. When benchmarking more than one device, a single problem size must used for a fair comparison.

The results you show for dot bandwidth on the GTX 1080 Ti look marginal between the default problem and anything larger. What is the difference?

psteinb:
> For both kernel types in CUDA we **have** to pick a "golden value" under the CUDA model. 1024 threads per thread block is such a number that could be classed as "best practice", which is what the focus of this code is: "If we follow best practices, how do we do?".

sure thing, but my point is that fixing the thread block size and the 
array size comes at the expense of a generic measurement, which the user 
of BabelStream needs to be aware of (and @ruyk already touched upon the 
issue of comparability with less low-level paradigms such as OpenMP, 
Sycl etc).
You can observe this in the plot I posted earlier: the default 
`array_size` is not yet at the saturation point (albeit it's close to 
it). So one has to wonder, how future GPU generations will behave.

> @psteinb STREAM is a memory bandwidth bound kernel, so I'm not sure I understand how you categorise it based on problem size, particularly with your reference to Roofline. The graph you plot is not Roofline as the axis are wrong. Roofline is FLOPS (y-axis) vs AI (x-axis), and the AI of the STREAM kernels are fixed no matter what problem size you run.

I guess , I wasn't too clear here with my wording. Sorry, to have caused 
confusion. I was mostly referring to the CUDA implementation of the dot 
kernel. Here, given you start to vary the thread block size, you can end 
up in the bandwidth saturated plateau (which is what BabelStream aims 
for I guess to make fair comparisons) or in the ramp before it (which is 
what you don't want to have I guess) - this behavior is what the blog 
post you mentioned illustrated IIRC.

FLOPS is just another bandwidth measure. the standard roofline plot 
settles on FLOPS vs. AI but does nothing else than illustrate/quantify 
the trade-off between memory bandwidth and compute bandwidth. you see 
this in the plot above as well. As long as you are on the ramp, you 
still have memory bandwidth that can be used. when you reach saturation, 
the only bound is frequency of the device, i.e. a metric related to 
computational performance (which is the root cause of my linguistic 
imprecision that I need to apologize for). So at the end of the day, I 
think the plot above and a classical roofline chart are very much 
related for GPUs. Even more so for the dot kernel (maybe not for the 
STREAM kernels as you pointed out correctly).

>
> The rules of original STREAM say the arrays have to be larger than last level cache. At one point we set the rules of BabelStream similarly (possibly in an old README but I can't find it right now). However on GPUs this is an awkward metric, and so when benchmarking one device in isolation you should increase the problem size until it changes little. When benchmarking more than one device, a single problem size must used for a fair comparison.
>
Yes, but on the GPU the situation is more severe as you have multiple 
hardware units (in nvidia lingo: the SMs) that each yield one or more 
channels to memory, i.e. only if you use all SMs, you will obtain the 
maximum memory bandwidth. which is (again), what my plot above 
illustrates. when the problem size is large enough, you can be sure, 
that all memory channels are used at maximum. This is my central concern 
with fixing the `array_size` AND `TBSIZE`.

> The results you show for dot bandwidth on the GTX 1080 Ti look marginal between the default problem and anything larger. What is the difference?
>
see my discussion above.

tomdeakin:
`array_size` is not fixed and can be set at runtime with the `--arraysize` flag; we've just set a default value that works OK for all devices we've tested to date. If new devices (of any kind) are released where this number is too small we will increase the default size.

In the CUDA code `TBSIZE` is fixed because the model requires a number to be chosen. Do you have a suggestion for choosing this number using the CUDA API that works well for all NVIDIA GPUs? Would this approach also viable in HC API?

Please state the numerical results for the GTX 1080 Ti for the default problem size and the larger problem size as it's hard to see what they are on your graph. I want to know what the percentage difference between the numbers are - are we talking a couple of %, or >10%?

psteinb:
on the `array_size` issue: good, this is what I've understood so far as well.

dynamically setting `TBSIZE`: don't know about this idea. As much as I'd like that to happen (you can infer the maximally supported thread block size from the CUDA API) I am not sure what it does to comparability to the results of other cards. Maybe this is a hen-and-egg problem. I need to think about that one.

not sure what you mean by "Please state the numerical results for the GTX 1080 Ti". the plot from above was for a GTX 1080 (no Ti involved). Here are the numerical values for completeness:

```
|-------------+--------+---------+--------------------|
|  array_size | sizeof | device  | dot_bw_mb_per_sec  |
|-------------+--------+---------+--------------------|
|  4096       | 4      | GTX1080 | 1249.209           |
|  8192       | 4      | GTX1080 | 2511.535           |
|  16384      | 4      | GTX1080 | 4998.551           |
|  32768      | 4      | GTX1080 | 9979.975           |
|  65536      | 4      | GTX1080 | 18388.328          |
|  131072     | 4      | GTX1080 | 37425.084          |
|  262144     | 4      | GTX1080 | 64513.859          |
|  524288     | 4      | GTX1080 | 108058.843         |
|  1048576    | 4      | GTX1080 | 145853.323         |
|  2097152    | 4      | GTX1080 | 191728.656         |
|  4194304    | 4      | GTX1080 | 217695.085         |
|  8388608    | 4      | GTX1080 | 232864.072         |
|  16777216   | 4      | GTX1080 | 242849.775         |
|  33554432   | 4      | GTX1080 | 245644.329         |
|  67108864   | 4      | GTX1080 | 247923.390         |
|  134217728  | 4      | GTX1080 | 248702.469         |
|  268435456  | 4      | GTX1080 | 249632.163         |
|  536870912  | 4      | GTX1080 | 249721.426         |
|-------------+--------+---------+--------------------|
```
the default array_size is missing out on 4GB/s  that is 1.6%. But this is one card only. Would be interesting to check this on a P100 for example or anything with much more DRAM than this 1080. As explained above, this might make the difference bigger. But maybe there is a misunderstanding here. I am not trying to convince you to increase the default `array_size` at all. I am mostly worried about the interplay of any `array_size` and `TBSIZE` etc.

So what do you suggest to do with this PR? From what I take from the discussion, I'd would play a bit with  `n_tiles` and see if I can find a value that maximizes the throughput. I'd make this a compile-time constant than similar to `TBSIZE`. What do you think?

tomdeakin:
Where values need to be hard coded in just pick something sensible, but code them in a way that they can be changed, probably with a `#define`.

tomdeakin:
I also just noticed you were running using single precision on the GTX1080. The default is double, which is what I thought you were running. I wonder if the default problem size in double precision on that device still shows a small difference from the maximum.

psteinb:
after quite some time, I got around to looking into changing the tile size a bit for the `dot` kernel. It turns out, changes don't have such a strong impact. therefor, I left the tile size configurable as a preprocessor parameter. I tested everything with rocm 1.6-77 on a R9 Nano. the results look ok:

```
GPU-STREAM
Version: 3.1
Implementation: HC
Running kernels 100 times
Precision: double
Using HC device AMD HSA Agent gfx8032
Copy        382799.207  0.00140     0.00155     0.00148     
Mul         391215.510  0.00137     0.00159     0.00146     
Add         351121.687  0.00229     0.00243     0.00238     
Triad       345814.135  0.00233     0.00242     0.00237     
Dot         271971.080  0.00197     0.00244     0.00206
```

Not sure if @AlexVlx has an opinion about it. I know that the ROCm team use babelstream within their benchmark suite.

psteinb:
any ideas about this PR?

tomdeakin:
Can you also pull out `n_tiles` to a preprocessor macro? Have a look at CUDAStream.h for inspiration.

psteinb:
done.

tomdeakin:
Thanks for contributing this. I've merged it.

Issue #20, opened 2016-12-13
Dot verification fails with single precision


jrprice:
We probably just need to increase the tolerance. The error will also be proportional to the size of the arrays (unlike with the other kernels), so we need to make sure whatever error checking tolerance we use is robust enough to avoid these sorts of false positives for any sort of input.

```
Validation failed on sum. Error 0.000209808
Sum was 39.7910385131836 but should be 39.7912483215332
```

tomdeakin:
We currently check that the `sum` array is within 1.0E-8 of the expected value for doubles and floats. We could either:

1. Use 1.0E-5 for for floats and 1.0E-8 for doubles
2. Factor in the array size somehow

Option 1 is simple but might hide errors. If the arrays contain correct values, then as long as the reduction is close for this benchmark it might be suitable.
Option 2 might be hard to quantify how we bias the array size.

Srinivasuluch:
Would it be possible to make the "sum" has double datatype irrespective of input args, "double or float" so that it gives accurate results.
The "sum" is an user data type used for comparison with "glodSum" value so it should not be matter.
I mean, change the "Template sum" to  "double sum"

tomdeakin:
For devices which do not support double precision would this not pose a problem?

zyzzyxdonta:
Hi,
is this issue still being worked on?

tomdeakin:
Yes, but we've not come up with a satisfactory solution yet.

zyzzyxdonta:
Thanks for your reply. Am I right in assuming that despite the verification failing, my measurements are still valid?

tomdeakin:
If it's just the reduction that fails (dot), and the other kernels are OK then the contents of the arrays should be correct. If the result is close enough on inspection but fails because of the tolerance then it's probably fine. If the result is 0.0 or some other nonsense number then it might have done something really wrong...

zyzzyxdonta:
Alright, thanks a lot!

tomdeakin:
@zjin-lcf suggested using different tolerances for the reduction result based on the data type (option 1 above).

tomdeakin:
Whilst reviewing #186 we discussed the fact that the `goldSum` value is computed numerically exactly using multiplication rather than repeated addition. I wonder if there is a way to get error bounds on the difference in algorithm from the FP rules.

PR #34, opened 2017-06-20
Add MPI to driver code

tomdeakin:
Add the ability to run BabelStream across multiple nodes with MPI. The input size is assumed to be **per MPI rank**, and therefore is similar to weak scaling. The reported bandwidth is of the entire parallel run.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -234,6 +234,11 @@ if (rank == 0)
 
   // Check solutions
   stream->read_arrays(a, b, c);
+
+#ifdef USE_MPI
+  // Only check solutions on the master rank in case verificaiton fails
+  if (rank == 0)
+#endif
   check_solution<T>(num_times, a, b, c, sum);
 
   // Display timing results
--- a/main.cpp
+++ b/main.cpp
@@ -276,18 +276,18 @@ if (rank == 0)
     double average = std::accumulate(timings[i].begin()+1, timings[i].end(), 0.0);
 
 #ifdef USE_MPI
-  // Collate timings
-  if (rank == 0)
-  {
-    MPI_Reduce(MPI_IN_PLACE, &min, 1, MPI_DOUBLE, MPI_MIN, 0, MPI_COMM_WORLD);
-    MPI_Reduce(MPI_IN_PLACE, &max, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
-  }
-  else
-  {
-    MPI_Reduce(&min, NULL, 1, MPI_DOUBLE, MPI_MIN, 0, MPI_COMM_WORLD);
-    MPI_Reduce(&max, NULL, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
-  }
-  sizes[i] *= procs;
+    // Collate timings
+    if (rank == 0)
+    {
+      MPI_Reduce(MPI_IN_PLACE, &min, 1, MPI_DOUBLE, MPI_MIN, 0, MPI_COMM_WORLD);
+      MPI_Reduce(MPI_IN_PLACE, &max, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
+    }
+    else
+    {
+      MPI_Reduce(&min, NULL, 1, MPI_DOUBLE, MPI_MIN, 0, MPI_COMM_WORLD);
+      MPI_Reduce(&max, NULL, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
+    }
+    sizes[i] *= procs;
 #endif
 
     average = average / (double)(num_times - 1);
--- a/main.cpp
+++ b/main.cpp
@@ -123,9 +123,19 @@ if (rank == 0)
   {
   std::streamsize ss = std::cout.precision();
   std::cout << std::setprecision(1) << std::fixed
-    << "Array size: " << ARRAY_SIZE*sizeof(T)*1.0E-6 << " MB"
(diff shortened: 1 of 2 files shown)


main.cpp
```diff
@@ -68,29 +90,45 @@ int main(int argc, char *argv[])
...
+#endif
+  {
   std::streamsize ss = std::cout.precision();
   std::cout << std::setprecision(1) << std::fixed
     << "Array size: " << ARRAY_SIZE*sizeof(T)*1.0E-6 << " MB"
     << " (=" << ARRAY_SIZE*sizeof(T)*1.0E-9 << " GB)" << std::endl;
   std::cout << "Total size: " << 3.0*ARRAY_SIZE*sizeof(T)*1.0E-6 << " MB"
     << " (=" << 3.0*ARRAY_SIZE*sizeof(T)*1.0E-9 << " GB)" << std::endl;
   std::cout.precision(ss);
+  }
```
jrprice: Maybe worth making these include the phrase "per rank" for these printouts when in MPI mode to make things crystal clear?

Could also show the actual total size, but not sure how interesting that is.

main.cpp
```diff
@@ -179,6 +237,10 @@ void run()
   check_solution<T>(num_times, a, b, c, sum);
```
jrprice: I think this needs to be in an `#ifdef USE_MPI` section as well - otherwise you get massively spammed if verification fails.

On a related note, we really need to sort out verification for the reduction as per #20, since this almost always fails now when using a large number of MPI ranks.

main.cpp
```diff
@@ -187,6 +249,7 @@ void run()
     << std::left << std::setw(12) << "Average" << std::endl;
 
   std::cout << std::fixed;
+  }
```
jrprice: This is fine, but might be interesting to see these results per rank *as well* as the aggregate bandwidth. This would make it easier to see how well things scale when running on lots of nodes.

main.cpp
```diff
@@ -201,19 +264,42 @@ void run()
...
     // Get min/max; ignore the first result
     auto minmax = std::minmax_element(timings[i].begin()+1, timings[i].end());
+    double min = *minmax.first;
+    double max = *minmax.second;
 
     // Calculate average; ignore the first result
-    double average = std::accumulate(timings[i].begin()+1, timings[i].end(), 0.0) / (double)(num_times - 1);
+    double average = std::accumulate(timings[i].begin()+1, timings[i].end(), 0.0);
+
+#ifdef USE_MPI
```
jrprice: Indentation inside this block looks a little off.

tomdeakin:
Re the Makefile changes. I guess the ideal solution will depend on how the changes are required. The `MPI=yes` option looks nice rather than having a separate build option, however the latter will be better if substantial changes are required for some models. I guess I'll just try to fold in the `MPI=yes` option and see how far I get. We can always change later...

main.cpp
tomdeakin: Is there a nice way to alter the string depending on the `USE_MPI` preprocessor define?

main.cpp
tomdeakin: Verification printing fixed in 5a64ce1.
#20 still needs to be addressed, although all this MPI code assumes `T` is `double` as I've used `MPI_Double` for messages. We should probably support `MPI_Float` too.

main.cpp
tomdeakin: One could always run the benchmark with `-np 1` (or run the non-MPI benchmark). We *could* also run the benchmark on a singe node first, and then print out the efficiency. This would probably need a bit of an overhaul in the way the driver works though. Not sure if it's worth the pain.

main.cpp
tomdeakin: Fixed in a57fdecc65f4ad8b847684bd332641c4af7c5123

tomdeakin:
ef22cbcc84c21790fdef61816131a46f0804f002 implements `MPI=yes` for the CUDA implementation. It builds the implementation to an object file, and then compiles this in with the main driver. If `MPI=yes` is set on the `make` invocation then the compiler switches out to `mpicxx` and adds the `USE_MPI` define to `EXTRA_FLAGS`.

Does this approach look OK? If so, I'll start to roll out similar changes to the other models.

tomdeakin:
Actually, piggybacking on `EXTRA_FLAGS` doesn't work if it's also set on the CLI... Need to rethink this!

main.cpp
jrprice: Not really, I think this is about as good at can get:

    std::cout << std::setprecision(1) << std::fixed
    #ifdef USE_MPI
      << "Array size (per MPI rank): "
    #else
      << "Array size: "
    #endif
      << "...";

main.cpp
jrprice: I just meant that it might be nice to print out the average per-rank bandwidths, I'm not talking about actually re-running on a single node.

e.g. If I run on a single P100 I get 550 GB/s. When I run on a cluster of 2000 P100s, I get some very large number, but do I still get 550 GB/s per rank?

To be honest I'm just being lazy - it's trivial for the user to open a calculator and do that division themselves. Feel free to ignore this comment :-)

jrprice:
> Does this approach look OK? If so, I'll start to roll out similar changes to the other models.

Yeah looks fine to me, bar the `EXTRA_FLAGS` issue you raise. I guess it'll just need an extra internal flags variable, which is annoying.

main.cpp
tomdeakin: Fixed in cb92f94b17df16b59d1ff0be8f11fba99da391bb

tomdeakin:
Closing as we'd prefer standalone MPI implementation. Opening new issue to this effect.

PR #36, opened 2017-07-27
Updated the Makefile for SYCL 

wyaneva:
This pull request contains:

-  An update to SYCL.make to use a new variable, containing the path to the ComputeCpp installation. The variable is `COMPUTECPP_PACKAGE_ROOT_DIR`, matching the one used in the build for [ComputeCpp SDK](https://github.com/codeplaysoftware/computecpp-sdk).

### DIFF
--- a/SYCLStream.cpp
+++ b/SYCLStream.cpp
@@ -49,9 +49,9 @@ SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
 
   queue = new cl::sycl::queue(dev, [&](cl::sycl::exception_list l) {
       try {
-      for(auto e: l) {
-        std::rethrow_exception(e);
-      }
+        for(auto e: l) {
+          std::rethrow_exception(e);
+        }
       } catch (cl::sycl::exception e) {
         std::cout << e.what();
       }
--- a/SYCLStream.cpp
+++ b/SYCLStream.cpp
@@ -47,16 +47,31 @@ SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
   std::cout << "Driver: " << getDeviceDriver(device_index) << std::endl;
   std::cout << "Reduction kernel config: " << dot_num_groups << " groups of size " << dot_wgsize << std::endl;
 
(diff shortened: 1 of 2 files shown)


SYCLStream.cpp
```diff
@@ -47,7 +47,15 @@ SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
...
   std::cout << "Reduction kernel config: " << dot_num_groups << " groups of size " << dot_wgsize << std::endl;
 
-  queue = new cl::sycl::queue(dev);
+  queue = new cl::sycl::queue(dev, [&](cl::sycl::exception_list l) {
+      try {
+        for(auto e: l) {
+          std::rethrow_exception(e);
+        }
+      } catch (cl::sycl::exception e) {
+        std::cout << e.what();
```
jrprice: Would all exceptions be fatal? If so, I think we should exit/abort/rethrow here, otherwise the application will try and continue after printing the `sycl::exception`.

SYCLStream.cpp
```diff
@@ -47,7 +47,15 @@ SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
   std::cout << "Driver: " << getDeviceDriver(device_index) << std::endl;
   std::cout << "Reduction kernel config: " << dot_num_groups << " groups of size " << dot_wgsize << std::endl;
 
-  queue = new cl::sycl::queue(dev);
+  queue = new cl::sycl::queue(dev, [&](cl::sycl::exception_list l) {
+      try {
```
jrprice: For consistency with the rest of the code, these opening braces should be on newlines.

wyaneva: I have moved the braces to new lines. b8f7a5427e82068f61b43373a8c59c6111858b35

SYCLStream.cpp
wyaneva: This is a fair point. The code now prints all SYCL exceptions and then throws an `std::runtime_exception`. b8f7a5427e82068f61b43373a8c59c6111858b35

jrprice:
Thanks for the contribution!

PR #37, opened 2017-08-02
Implement --triad-only switch

jrprice:
For measuring cache bandwidth.


psteinb:
Can you elaborate on the relation of the triad kernel to the cache bandwidth? (i.e. in comparison to `add`)

jrprice:
There's nothing particularly special about triad; `add` would do the job just fine as well. The point here is just to run a single function, and we picked triad because it is typically the one used when reporting STREAM bandwidth numbers.

psteinb:
Interesting. thanks for the heads-up.

PR #39, opened 2017-12-08
SYCL implementation adapted to 1.2.1 interface

antonrv:
Changes in SYCLStream to adapt it to the new 1.2.1 SYCL interface
https://www.khronos.org/news/press/the-khronos-group-releases-finalized-sycl-1.2.1


Ruyk:
These changes make ComputeCpp CE 0.5 work with Babel Stream. Not sure if you guys want to check for language version and keep the old code too? Neither ComputeCpp or TriSYCL will support the old interface AFAIK, changes are sufficiently minor to not be a problem for users...

tomdeakin:
Thanks for the update! @jrprice tested this on AMD and worked for us, so I've merged the commit.

PR #45, opened 2018-02-15
Check sum with 2% error instead of abs error.

### LINKED ISSUE #20: Dot verification fails with single precision

tomdeakin:
This seems to help validate single precision runs where the sum
is as close as we can expect for these large arrays.

This fixes #20.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -425,10 +425,10 @@ void check_solution(const unsigned int ntimes, std::vector<T>& a, std::vector<T>
     std::cerr
       << "Validation failed on c[]. Average error " << errC
       << std::endl;
-  // Check sum to 8 decimal places
-  if (!triad_only && errSum > 1.0E-8)
+  // Check percentage error of sum. Should be within 2%.
+  if (!triad_only && errSum/goldSum > 2.0E-2)
     std::cerr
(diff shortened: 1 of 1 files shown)


jrprice:
This is extremely sensitive to not just array size, but also number of threads (for OpenMP).

For example, running with the default array size with just 2 threads using the Intel compiler produces an error a little over 2%.

For a single thread running with 1GB arrays (2^28) elements, I'm getting an error of 44%, which is worryingly large.

That said, I'm not sure there's a better solution. We may get lower errors by tweaking the initial/scalar constants to keep the dot result from getting too large (where the coarseness of the floating point representation becomes an issue)?

tomdeakin:
Oh wow, those are big errors. I wonder if we should normalise the dot product by the array size just in the checking script. Maybe that would net out a factor of the array size?

tomdeakin:
Closing as this doesn't sufficiently solve Issue #20

Issue #48, opened 2018-03-16
Bablestream is not working with HIP programming model on the latest ROCm1.7.1(1.7-137)


rkothako:
Bablestream is not working with HIP programming model on the latest ROCm1.7.1(1.7-137).
Application should be modified as per the latest changes in HIP, correct?

Can you please take a look at below log and provide the resolution?

taccuser@ROCM-DTB-VG10:~/Desktop/BabelStream$ make -f HIP.make
/opt/rocm/hip/bin/hipcc -std=c++11 -DHIP main.cpp HIPStream.cpp -o hip-stream
HIPStream.cpp:129:3: error: no matching function for call to 'hipLaunchKernel'
hipLaunchKernel(HIP_KERNEL_NAME(copy_kernel), dim3(array_size/TBSIZE), ...
^~~~~~~~~~~~~~~
HIPStream.cpp:270:16: note: in instantiation of member function
'HIPStream::copy' requested here
template class HIPStream;
^
/opt/rocm/hip/include/hip/hcc_detail/functional_grid_launch.hpp:142:6: note:
candidate function [with Args = <float *, float >, F = void
()(hip_impl::Empty_launch_parm, float *, float )] not viable: no overload
of 'copy_kernel' matching 'void ()(hip_impl::Empty_launch_parm, float *,
float *)' for 1st argument
void hipLaunchKernel(
^
HIPStream.cpp:162:3: error: no matching function for call to 'hipLaunchKernel'
hipLaunchKernel(HIP_KERNEL_NAME(add_kernel), dim3(array_size/TBSIZE), d...
^~~~~~~~~~~~~~~
HIPStream.cpp:270:16: note: in instantiation of member function
'HIPStream::add' requested here
template class HIPStream;
^
/opt/rocm/hip/include/hip/hcc_detail/functional_grid_launch.hpp:142:6: note:
candidate function [with Args = <float *, float *, float >, F = void
()(hip_impl::Empty_launch_parm, float *, float *, float )] not viable: no
overload of 'add_kernel' matching 'void ()(hip_impl::Empty_launch_parm,
float *, float *, float *)' for 1st argument
void hipLaunchKernel(
^
HIPStream.cpp:146:3: error: no matching function for call to 'hipLaunchKernel'
hipLaunchKernel(HIP_KERNEL_NAME(mul_kernel), dim3(array_size/TBSIZE), d...
^~~~~~~~~~~~~~~
[...]

PR #49, opened 2018-03-19
Make hip-stream buildable with latest version of HIP

mangupta:
This addresses the issue mentioned in #48


tomdeakin:
Thanks for the contribution!

PR #51, opened 2018-05-02
Minor SYCL and OpenCL fixes

Ruyk:
* Fix for OpenCL/SYCL behaviour with global objects
* Separated makefile for SYCL to get more concise error messages and reduce compilation time


tomdeakin:
Thanks for this, sorry it's taken a while to merge in.

PR #56, opened 2019-04-09
Add a --mibibytes flag to output bandwidth and array sizes in base 2

### LINKED ISSUE #55: Add option for base 2 MiB/s output
BabelStream uses base 10 output where MB = 10^6.

It would be useful to add a `--base2` or `--mibibytes` option to use base 2 output to calculate bandwidth for MiB = 2^20.

tomdeakin:
This sets MiB = 2^20, GiB = 2^30 rather than the default of
MB = 10^6 and GB = 10^9.

This would fix #55 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 - OpenMP GNU compiler now uses native target flag.
 - Support CSV output for Triad only running mode.
 - NEC and PGI compiler option for OpenMP version.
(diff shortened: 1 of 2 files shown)


patrickatkinson:
This all looked fine except for the printing of the bandwidth to the command line, which still only used 10^6. 

My commit should fix this (https://github.com/UoB-HPC/BabelStream/pull/56/commits/c50eba9caf1bc1073c59f4f15317f2d81509fd8b).

PR #57, opened 2019-05-22
Fixed compatibility issue with SYCL-1.2.1

tianshilei1992:
Current implementation is not compatible with latest SYCL-1.2.1. This patch is to make it work.

### DIFF
--- a/SYCLStream.cpp
+++ b/SYCLStream.cpp
@@ -104,8 +104,7 @@ void SYCLStream<T>::copy()
   {
     auto ka = d_a->template get_access<access::mode::read>(cgh);
     auto kc = d_c->template get_access<access::mode::write>(cgh);
-    cgh.parallel_for<copy_kernel>(p->get_kernel<copy_kernel>(),
-          range<1>{array_size}, [=](item<1> item)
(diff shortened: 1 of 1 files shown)


tomdeakin:
@Ruyk, does this update look consistent with what is supported by the ComputeCpp compiler? I believe that supports SYCL 1.2.1 now.

tianshilei1992:
@tomdeakin ComputeCpp claims that they're fully SYCL 1.2.1 compatible. :)

GeorgeWeb:
Hello,

I am aware that I made a PR which is a duplicate of this one without realizing that this one exists in the first place when I went and pushed mine.

**However**, I have given a complete explanation on why are the changes needed at this stage, putting some light on this **non-conformant (yet)** functionality that is however supported by **ComputeCpp**.

I suggest going over it before deciding how to proceed: **[PR Link](https://github.com/UoB-HPC/BabelStream/pull/58).**

Regards,

Georgi

Ruyk:
The parallel_for overload that takes a kernel as a first argument was in SYCL 1.2 but somehow we missed it on the 1.2.1 spec. In my opinion it was just an oversight, we've raised this internally in the Khronos group.

tomdeakin:
Fixed by #58.

PR #58, opened 2019-06-19
Making BabelStream's SYCL code compliant

GeorgeWeb:
**1. Deprecated code**

```
'get_global' is deprecated: SYCL 1.2.1
revision 3 replaces nd_item::get_global with nd_item::get_global_id.
```

```
'get_local' is deprecated: SYCL 1.2.1 revision 3 replaces
nd_item::get_local with nd_item::get_local_id.
```
 

**2. Non-official code**

Before closing this with saying "ComputeCpp claims it fully conformant with SYCL 1.2.1", please note that while this is correct, the notion of using the pre-built kernel in with `parallel_for` is _not currently_ part of the SYCL Specification - v1.2.1 Revision 5.

**The issue:**
The SYCL specification does not include an overload of `handler::parallel_for` that accepts syclKernel built from a program. For example, `parallel_for(syclKernel, nd_range, functor)` is not conformant with specification at the moment. The reason because of which this **works with ComputeCpp** for quite some time now, is that this functionality _has probably been_ a part of past version of the specification.

**Suggestion:**
Thus, I suggest disabling/removing the use of this feature from BabelStream just in case it needs to be fully SYCL compliant and independent to a specific SYCL implementation (i.e., triSYCL doesn't support it). **Eventually**, when/if the feature gets added officially to the specification, it can be re-introduced in the code.

---

**Reference:**
All function overloads for `parallel_for` can be seen in the latest revision of the official SYCL spec:
https://www.khronos.org/registry/SYCL/specs/sycl-1.2.1.pdf#page=167

### DIFF
--- a/SYCLStream.cpp
+++ b/SYCLStream.cpp
@@ -11,14 +11,11 @@
 
 using namespace cl::sycl;
 
-
 // Cache list of devices
 bool cached = false;
 std::vector<device> devices;
 void getDeviceList(void);
-#ifdef COMPUTECPP_CE
 program * p;
-#endif
 
 template <class T>
 SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
@@ -70,17 +67,6 @@ SYCLStream<T>::SYCLStream(const unsigned int ARRAY_SIZE, const int device_index)
     }
   });
   
-  #ifdef COMPUTECPP_CE
-  /* Pre-build the kernels */
-  p = new program(queue->get_context());
-  p->build_with_kernel_type<init_kernel>();
-  p->build_with_kernel_type<copy_kernel>();
-  p->build_with_kernel_type<mul_kernel>();
-  p->build_with_kernel_type<add_kernel>();
-  p->build_with_kernel_type<triad_kernel>();
-  p->build_with_kernel_type<dot_kernel>();
-  #endif
-
   // Create buffers
   d_a = new buffer<T>(array_size);
   d_b = new buffer<T>(array_size);
@@ -95,9 +81,7 @@ SYCLStream<T>::~SYCLStream()
   delete d_b;
   delete d_c;
   delete d_sum;
-  #ifdef COMPUTECPP_CE
   delete p;
-  #endif
   delete queue;
   devices.clear();
 }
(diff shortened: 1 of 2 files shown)


Ruyk:
I suggest ifdef out that part of the code with the COMPUTECPP_CE macro, so that it continues to work in ComputeCpp until there is a resolution to the issue in the Khronos SYCL group.

GeorgeWeb:
Yeah, that would make for a clearer and generally better temporary solution than complete removal of the feature. Thanks for the jumping into the discussion @Ruyk !

tomdeakin:
Thanks for this discussion. I can't remember exactly the reason we used them here, but I have a vague memory that it was around the need to template on the data type (`double` and `float`).

We try to demonstrate "best practice" SYCL with BabelStream, so is the use of these pre-built kernels recommended still, or should we just remove this entirely?

GeorgeWeb:
Performance-wise I would prefer it, especially in the case of benchmarking, where the comparison against other models should be as fair as possible. 

With the `parallel_for(nd_range, functor)` approach, you would get some overhead in the first execution as the kernel will be built on initial queue submission. However, normally I would usually go with this default approach because I find it safer and more clear.

In the case of BabelStream I would certainly stick with the use of pre-build the kernels for more accurate (that was the word, not "fair") benchmarking.

Ruyk:
Performance-wise, definitely go for pre-built kernels. It can't harm any implementation, and will avoid unnecessary overhead on implementations that rely on online compilation of kernels.

tomdeakin:
Thanks @Ruyk and @GeorgeWeb. BabelStream, like STREAM, doesn't count the first iteration in the timing, so any first step overhead wouldn't be reported. Would your opinion change for a larger, more production style codebase?

GeorgeWeb:
Not really. Generally, if I can and want to make sure I avoid any unnecessary overhead, as mentioned by Ruyman, I will still pre-build the kernels.

GeorgeWeb:
Right, more on this, after a conversation with @Ruyk 

After the first build of the kernel the binaries are cached so as long as the `context` stays the same, there should not be any significant (if any at all) performance difference, especially if BabelStream doesn't take the first iteration into an account. 

In our discussion we agreed on that, considering that we measured performance with both approaches, it may be best to remove the use program and pre-built kernels in the code. It is more readable and still as performant.

tomdeakin:
Thanks for this @GeorgeWeb and @Ruyk, this looks great. 😄

PR #62, opened 2020-03-12
Update cl2.hpp to handle OpenCL platforms with zero devices.

Noerr:
Need to more carefully handle the case of a platform detected with zero devices.  In the first call to clGetDeviceIDs, this will return code CL_DEVICE_NOT_FOUND and set n=0 if no devices are reported by the platform. ( https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clGetDeviceIDs.html )
The second call to clGetDeviceIDs cannot be allowed occur if n=0, otherwise CL_INVALID_VALUE is reported.
A platform with zero devices is not an error condition.  Device enumeration should simply move on.

### DIFF
--- a/CL/cl2.hpp
+++ b/CL/cl2.hpp
@@ -2346,15 +2346,17 @@ public:
             return detail::errHandler(CL_INVALID_ARG_VALUE, __GET_DEVICE_IDS_ERR);
         }
         cl_int err = ::clGetDeviceIDs(object_, type, 0, NULL, &n);
-        if (err != CL_SUCCESS) {
+        if (err != CL_SUCCESS && err != CL_DEVICE_NOT_FOUND) {
             return detail::errHandler(err, __GET_DEVICE_IDS_ERR);
         }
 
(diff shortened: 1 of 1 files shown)


tomdeakin:
The `cl2.hpp` header in this repository is a local clone of the [official Khronos OpenCL C++ header](https://github.com/KhronosGroup/OpenCL-CLHPP). We periodically copy it from upstream into this repository as it does not always ship with OpenCL platforms.

I'd recommend you open an Issue against the [header source](https://github.com/KhronosGroup/OpenCL-CLHPP) so that we can pick it up from there.

Noerr:
The proposed change is now merged into the upstream master. https://github.com/KhronosGroup/OpenCL-CLHPP/pull/85
@tomdeakin I recommend updating the cl2.hpp copy in this project.

tomdeakin:
@Noerr Done! Thanks for the contribution, and I'm really pleased it went into the official header quickly.

PR #66, opened 2020-05-21
Add GNU OpenACC support for AMD GCN

ams-cs:
This adds GNU OpenACC support, similar to that recently added for OpenMP. There's also a few code changes needed to support GNU C++ and AMD devices.

* Autodetect the device type, rather than hard-code NVidia.
* Add GNU command line options to the makefile, and adjust the "restrict"
extension usage. For now, we assume the toolchain is only configured for one
accelerator.


tomdeakin:
Thanks a lot for the PR @ams-cs, much appreciated.

Issue #64, opened 2020-05-11
Add build rules for hipSYCL and DPC++ to SYCL version


tomdeakin:

tomdeakin:
# hipSYCL CUDA
```syclcc-clang -std=c++17 -DSYCL -O3 main.cpp SYCLStream.cpp --hipsycl-gpu-arch=sm_75 --hipsycl-platform=cuda```

# hipSYCL AMD
```syclcc-clang -DSYCL -O3 main.cpp SYCLStream.cpp --hipsycl-gpu-arch=gfx906 --hipsycl-platform=rocm```

tomdeakin:
Done in #79

PR #69, opened 2020-06-01
[SYCL] Makefile to build SYCL variant with DPCPP

Ruyk:
* Requires DPCPP_ROOT_DIR to be defined
* Target triple hardcoded to CUDA

### DIFF
--- a/DPCPP.make
+++ b/DPCPP.make
@@ -1,0 +1,16 @@
+
+DPCPP_FLAGS=-fsycl -fsycl-targets=nvptx64-nvidia-cuda-sycldevice -fsycl-unnamed-lambda
+
+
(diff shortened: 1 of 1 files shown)


tomdeakin:
Thanks for this, but ideally (#64) we'd want to have a `SYCL.make` in the style of our `OpenMP.make` file. This means we could pick DPC++ or hipSYCL with via a `COMPILER` variable, and then target the different backends with the `TARGET` variable.

The build systems are pretty consistent between these two compilers so it should be fairly easy to do that. It would be even better if we could have ComputeCpp as an option in the same Makefile too, but the split invocation might not make this easy to do.

tomdeakin:
#79 should have generalised this for three SYCL compilers.

PR #88, opened 2021-02-04
Add PRK Nstream kernel

tomdeakin:
This PR adds a new kernel to BabelStream, the `nstream` kernel borrowed from [PRK](https://github.com/ParRes/Kernels).
The new kernel computes:
```
a[i] += b[i] + scalar * c[i];
```
This has 3 reads and 1 write which is a heaver read ratio than Triad. Non-temporal stores on CPUs are also not required as the `a` array must be first read and be placed in cache.

~To support this, the initial values of the arrays are changed to ensure the results validate.~ no longer required.

The OpenACC kernel does not seem to respect the `+=`; it does not include the initial value of `a` just overwrites it the same as Triad which is incorrect.

The modern C++ is hard to test because not all compilers support the required APIs in the Parallel STL, or else the software stack is fragile.

### DIFF
--- a/ACCStream.cpp
+++ b/ACCStream.cpp
@@ -121,6 +121,22 @@ void ACCStream<T>::triad()
 }
 
 template <class T>
+void ACCStream<T>::nstream()
+{
+  const T scalar = startScalar;
+
+  int array_size = this->array_size;
+  T * restrict a = this->a;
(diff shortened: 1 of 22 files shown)


tomdeakin:
The kernel has now been added in isolation as it is too difficult to maintain stable input values.

PR #91, opened 2021-03-05
CMake+CI integration

### LINKED ISSUE #38: Compile errors with RAJA 0.3.x
RAJA 0.3.x has breaking changes which are not backwards compatible, which means that BabelStream no longer builds with the latest versions of RAJA.

tom91136:
This PR adds CMake support for BabelStream.

More and more userland parallel frameworks are moving to CMake and dropping Makefiles entirely. Even
if they do support Makefiles, those typically aren't up-to-date so newer versions may not work as
expected. As Babelstream is intended to compile on multiple compilers, we also can also take
advantage of CMake's built-in support for platform and compiler detection which eliminates many
repetitive snippets in existing Makefiles.

### CMake

README has been updated with up-to-date CMake instructions.

A few quality of life improvements are listed here:

* The HC implementation has been moved to the `legacy` directory as AMD has deprecated the HCC
  compiler.
* Passing additional flags can now be delimited with `;`, (
  i.e `-DCXX_EXTRA_FLAGS="-VERY=SLOW;-O0;-march=pentium2"`). This has the advantage of avoiding
  word-splitting issues in shell scripts.
* All the required and optional flags are indicated upfront, for example:
  ```shell
  -- Common Release flags are `-O3`, set RELEASE_FLAGS to override
  -- CXX_EXTRA_FLAGS: 
          Appends to common compile flags, this will be appended at link phase as well.
          To use separate flags at link phase, set `CXX_EXTRA_LINKER_FLAGS`
  -- CXX_EXTRA_LINK_FLAGS: 
          Appends to link flags (i.e appended only at link phase) which appears *before* the objects.
          Do not use this for linking libraries as the link line is order dependent
  -- CXX_EXTRA_LIBRARIES: 
          Append to link flags which appears *after* the objects, use this for linking extra libraries (e.g `-lmylib`, or simply `mylib`) 
  -- CXX_EXTRA_LINKER_FLAGS: 
          Append to linker flags (i.e GCC's -Wl or equivalent)
  -- Available models:  OMP;OCL;STD;STD20;HIP;CUDA;KOKKOS;SYCL;ACC;RAJA
  -- Selected model  :  CUDA
  -- Supported flags:
  
     CMAKE_CXX_COMPILER (optional, default=c++): Any CXX compiler that is supported by CMake detection, this is used for host compilation
     MEM (optional, default=DEFAULT): Device memory mode:
          DEFAULT   - allocate host and device memory pointers.
          MANAGED   - use CUDA Managed Memory.
          PAGEFAULT - shared memory, only host pointers allocated.
     CMAKE_CUDA_COMPILER (required): Path to the CUDA nvcc compiler
     CUDA_ARCH (required): Nvidia architecture, will be passed in via `-arch=` (e.g `sm_70`) for nvcc
     CUDA_EXTRA_FLAGS (optional, default=): Additional CUDA flags passed to nvcc, this is appended after `CUDA_ARCH`
  
  -- Picked up model specific flags for this build:
  
     CMAKE_CXX_COMPILER = `/usr/bin/clang++`
     MEM = `PAGEFAULT`
     CMAKE_CUDA_COMPILER = `/home/tom/Downloads/nvhpc_2021_212_Linux_x86_64_cuda_11.2/install_components/Linux_x86_64/21.2//cuda/11.2/bin/nvcc`
     CUDA_ARCH = `sm_70`
     CUDA_EXTRA_FLAGS = ``
  ```

* This PR also includes a copy of
  the [official OpenCL headers](https://github.com/KhronosGroup/OpenCL-Headers) that has been tested
  for compatibility. This is required because ComputeCpp's OCL header requirement is a bit
  unforgiving so headers that are too old causes compile errors. The OCL model also defaults to this
  header for convenience.

* This PR also removes all hardcoded use of platform dependent build flags (e.g `-march=native`). It
  is expected that the user would append the appropriate flags for their target platform, as
  indicated in the build:
  ``` 
  Common Release flags are `-O3`, set RELEASE_FLAGS to override.
  ```
  The user could either set `RELEASE_FLAGS` or use `CXX_EXTRA_FLAGS`, for example:
  ```shell
  -DCXX_EXTRA_FLAGS=-Ofast;-march=native;-mtune=native
  ```
* As per the recommendation, this PR removes the need to install RAJA/Kokkos locally. Both builds
  are in-tree, meaning that the framework and BabelStream are compiled in the same build process.
  The flags are shared across the framework and BabelStream when appropriate.

While migrating, a few minor bugs was also fixed:

* The RAJA version used was old and the API has since changed, this PR re-aligned it with latest
  version (0.13.0), closes #38.
* RAJA didn't implement the new `nstream` kernel, this PR includes a dummy implementation so that it
  compiles; actual implementation to be addressed in another PR.
* `STDStream`'s constructor impl. has `noexcept` but not in the header, GCC doesn't like that.
* GCC can't handle ACC's offload pragmas when the referenced pointers are not in scope (even if they
  are in the same C++ class). The pointers were simply brought into scope before the pragmas as done
  in `OMPStream` with offload.

### CI

To give this PR the best possible chance of being merged, I've add CI via GitHub actions. The PR
attempts to test-compile all models with at least one compatible compiler. For sanity, we also
test-run the OMP version on GCC to make sure the host code isn't broken.

The CI run script also serves as a concrete example for what flags are required for each
model/compiler.

For now, I've included the following compilers:

* GCC 10
* Clang 9
* NVHPC 21.2
* AOCC 2.3.0
* AOMP 11.12
* HIPCC 4.0
* DPC++ 20210106
* hipSYCL 0.9.0

The CI gets triggered on commit and PR, the output and status can be found under GitHub's `Actions`
tab. Here's a snippet of the GCC CI pass:

```
[PASS!](OMP->./build/OMP_gcc_build): -DMODEL=OMP -DCMAKE_CXX_COMPILER=/usr/bin/g++-10
       CMAKE_CXX_COMPILER = `/usr/bin/g++-10`
    /usr/bin/g++-10 -DOMP  -DNDEBUG -O3 -fopenmp -std=c++17 -o CMakeFiles/babelstream.dir/OMPStream.cpp.o -c /home/runner/work/github_action_test/github_action_test/OMPStream.cpp
    /usr/bin/g++-10 -DOMP  -DNDEBUG -O3 -fopenmp -std=c++17 -o CMakeFiles/babelstream.dir/main.cpp.o -c /home/runner/work/github_action_test/github_action_test/main.cpp
    /usr/bin/g++-10  -DNDEBUG CMakeFiles/babelstream.dir/OMPStream.cpp.o CMakeFiles/babelstream.dir/main.cpp.o -o babelstream  /usr/lib/gcc/x86_64-linux-gnu/10/libgomp.so /usr/lib/x86_64-linux-gnu/libpthread.so 
    ././build/OMP_gcc_build/babelstream: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=06a6ef30d235652e7a7d4464c4583a3f8ae412be, not stripped
Sanity checking GCC OMP build...
BabelStream
Version: 3.4
Implementation: OpenMP
Running kernels 10 times
Precision: double
Array size: 8.4 MB (=0.0 GB)
Total size: 25.2 MB (=0.0 GB)
Copy        41080.051   0.00041     0.00048     0.00044     
Mul         41921.765   0.00040     0.00055     0.00044     
Add         49490.021   0.00051     0.00062     0.00056     
Triad       49509.493   0.00051     0.00060     0.00055     
Dot         27489.933   0.00061     0.00071     0.00065     
[PASS!](STD->./build/STD_gcc_build): -DMODEL=STD -DCMAKE_CXX_COMPILER=/usr/bin/g++-10 -DCXX_EXTRA_LIBRARIES=tbb
       CMAKE_CXX_COMPILER = `/usr/bin/g++-10`
    /usr/bin/g++-10 -DSTD  -DNDEBUG -O3 -std=c++17 -o CMakeFiles/babelstream.dir/STDStream.cpp.o -c /home/runner/work/github_action_test/github_action_test/STDStream.cpp
    /usr/bin/g++-10 -DSTD  -DNDEBUG -O3 -std=c++17 -o CMakeFiles/babelstream.dir/main.cpp.o -c /home/runner/work/github_action_test/github_action_test/main.cpp
    /usr/bin/g++-10  -DNDEBUG CMakeFiles/babelstream.dir/STDStream.cpp.o CMakeFiles/babelstream.dir/main.cpp.o -o babelstream  -ltbb 
    ././build/STD_gcc_build/babelstream: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=fb512b7ed589a472704daf7687cf8d2032037bce, not stripped
[PASS!](STD20->./build/STD20_gcc_build): -DMODEL=STD20 -DCMAKE_CXX_COMPILER=/usr/bin/g++-10 -DCXX_EXTRA_LIBRARIES=tbb
       CMAKE_CXX_COMPILER = `/usr/bin/g++-10`
    /usr/bin/g++-10 -DSTD20  -DNDEBUG -O3 -std=c++2a -o CMakeFiles/babelstream.dir/main.cpp.o -c /home/runner/work/github_action_test/github_action_test/main.cpp
    /usr/bin/g++-10 -DSTD20  -DNDEBUG -O3 -std=c++2a -o CMakeFiles/babelstream.dir/STD20Stream.cpp.o -c /home/runner/work/github_action_test/github_action_test/STD20Stream.cpp
    /usr/bin/g++-10  -DNDEBUG CMakeFiles/babelstream.dir/STD20Stream.cpp.o CMakeFiles/babelstream.dir/main.cpp.o -o babelstream  -ltbb 
    ././build/STD20_gcc_build/babelstream: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=709f1c0a23e14f6ab6b290a434223947c96f1415, not stripped
[PASS!](ACC->./build/ACC_amd_gcc_build): -DMODEL=ACC -DCMAKE_CXX_COMPILER=/usr/bin/g++-10 -DCXX_EXTRA_FLAGS=-foffload=amdgcn-amdhsa
       CMAKE_CXX_COMPILER = `/usr/bin/g++-10`
    /usr/bin/g++-10 -DACC -Drestrict=__restrict  -DNDEBUG -O3 -foffload=amdgcn-amdhsa -fopenacc -o CMakeFiles/babelstream.dir/main.cpp.o -c /home/runner/work/github_action_test/github_action_test/main.cpp
    /usr/bin/g++-10 -DACC -Drestrict=__restrict  -DNDEBUG -O3 -foffload=amdgcn-amdhsa -fopenacc -o CMakeFiles/babelstream.dir/ACCStream.cpp.o -c /home/runner/work/github_action_test/github_action_test/ACCStream.cpp
    /usr/bin/g++-10  -DNDEBUG -foffload=amdgcn-amdhsa -fopenacc CMakeFiles/babelstream.dir/ACCStream.cpp.o CMakeFiles/babelstream.dir/main.cpp.o -o babelstream 
    ././build/ACC_amd_gcc_build/babelstream: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=453672c3822633b65fbf2e91b3345426e9900f75, not stripped
```

The CI process first downloads and prepare all of compilers, the install directory is cached to disk
and compressed to save time for future builds. Finally, the source is tested with each compiler for
the applicable models. The whole process takes about 17 minutes to complete.

Once the compiler preparation has been cached, subsequent builds range from 10 to 12 minutes. We got
3000 minutes of CI time so this should be plenty based on our commit frequency.


tomdeakin:
This is really excellent @tom91136 ! Loads of great stuff here.

> This PR also removes all hardcoded use of platform dependent build flags (e.g -march=native). It
is expected that the user would append the appropriate flags for their target platform, as
indicated in the build:

We want to make sure the best flags are always used, so the defaults should contain the best flags for each compiler choice.
I think this change might make it too easy to forget to turn on `-march=native` or anything else that's needed, as the benchmark won't do this anymore. Is there a way to capture the recommended flags for each compiler/model combination and include them by default?

tom91136:
I was a bit hesitant to do this out of the box as we typically use the "compile on login node and submit" workflow. 

But anyway, right now `RELEASE_FLAGS` is set to `-O3`, we can append `-march=native` to that quite easily.
I think most of the compilers I've tested support `-O3`, let me make a commit to see if `-march=native` breaks anything in the CI.
If something does break, we can always make changes on a per compiler (`CMAKE_CXX_COMPILER_ID`) or a per model basis.

tom91136:
OK, `-march=native` seems fine by the looks of it.

tom91136:
No problem, I'm adding all the vendor specific flags back now.
CMake supports dynamic variable names like in Make so it's fairly straightforward. 
This will take a few more days as I'm double checking the more exotic ones such as NEC `nc++`, apparently CMake identify it as gcc 6.5.

tomdeakin:
Thanks for adding in all the flags @tom91136 

The binary name is now `babelstream` irrespective of what is built. Can we retain the unique binary names for each model?

I'm wondering about the strict requirement for C++17 in the Makefile. For most things, C++11 should be sufficient, with the exception of SYCL 2020 (C++17) and the C++ Parallel STL versions (C++17 and C++20). Doing the minimum (`cmake -DMODEL=OMP` .) will fail with the usual GCC 4.6.
Can we default to C++11, and then specialise to C++14/17 in the model files?

tom91136:
The C++17 requirement is only in OMP by mistake, I've dropped that requirement with the latest commit. Prefixed executables are also implemented and the CI seems happy.

tomdeakin:
Great - does CMake (or somewhere in the CMake) assume C++11? I think we require that for some of the standard library stuff in `main.cpp` like `<chrono>`.

tom91136:
Yeah, I'll set a min requirement for 11. Normally we want to use CMake's feature detection so compilers that implement parts of a standard can still work, like `target_compile_features(babelstream PUBLIC cxx_constexpr)`, I don't think that works for standard libraries so I'm going with `target_compile_features(babelstream PUBLIC cxx_std_11)` .

tomdeakin:
This is probably a daft CMake question, but why do it with a `target_compile_features` in some places and a `set(CMAKE_CXX_STANDARD ??)` in others?

tom91136:
`target_compile_features($EXE PUBLIC <feature>)` is usually a better way than just setting variables directly as it ties it to that particular executable.
In theory, we can have multiple executables with different features and cxx versions.
For CXX features and version, `cxx_std_??` is added only in CMake ~3.8 so we got a lot of momentum for `CMAKE_CXX_STANDARD`
After having a second look, I now remember why I used `CMAKE_CXX_STANDARD` in the first place: Kokkos and RAJA reads these variables. I'll just go with ye olde `set(CMAKE_CXX_STANDARD 11)`.

tom91136:
Looks like ROCm 4.1 was released and broke all the HIP related checks; apparently actual ROCm users from various distros are also reporting regression, let me see if i can just download a specific version of ROCm and save it to cache.

tom91136:
Gave up on finding a working binary distribution that we can just download and untar. Upgrading to 4.1.0 via apt for now.

andreipoe (approved):
I've cleaned up the messages a little, both in the README and in the CMake configuration. I was able to build and run the code following the instructions, so that's fine by me.

tomdeakin (approved):

PR #96, opened 2021-04-16
Include stdlib.h for aligned_alloc

tob2:
Silence "error: there are no arguments to 'aligned_alloc' that depend
on a template parameter, so a declaration of 'aligned_alloc' must be
available"

* OMPStream.cpp: #include <stdlib.h>.
* RAJAStream.cpp: Likewise.

### DIFF
--- a/OMPStream.cpp
+++ b/OMPStream.cpp
@@ -5,6 +5,7 @@
 // For full license terms please see the LICENSE file distributed with this
 // source code
 
+#include <cstdlib>  // For aligned_alloc
 #include "OMPStream.h"
 
 #ifndef ALIGNMENT
--- a/RAJAStream.cpp
+++ b/RAJAStream.cpp
@@ -5,6 +5,7 @@
 // For full license terms please see the LICENSE file distributed with this
 // source code
 
+#include <cstdlib>  // For aligned_alloc
 #include <stdexcept>
 #include "RAJAStream.hpp"
 


tob2:
@tomdeakin please review - thanks!

tomdeakin:
Thanks for the PR, @tob2.
Can you provide some more details about which compilers/systems you are using that give you the error?
I think we should include the header as `<cstdlib>` rather than `<stdlib.h>` though, which looks to be better practice.

tob2:
>  Can you provide some more details about which compilers/systems you are using that give you the error?

That was seen by colleague of mine with GCC 10 (x86-64-gnu-linux) with nvidia nvptx offloading configured:
```
$ x86_64-none-linux-gnu-g++ main.cpp -O3 -std=c++11 -march=native -DOMP OMPStream.cpp -DOMP_TARGET_GPU -fopenmp -foffload=nvptx-none -fdiagnostics-color=never -lm -o omp-stream
OMPStream.cpp: In constructor 'OMPStream<T>::OMPStream(int, int)':
OMPStream.cpp:20:17: error: there are no arguments to 'aligned_alloc' that depend on a template parameter, so a declaration of 'aligned_alloc' must be available [ESC]8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-fpermissive^G-fpermissiveESC]8;;^G]
   20 |   this->a = (T*)aligned_alloc(ALIGNMENT, sizeof(T)*array_size);
      |                 ^~~~~~~~~~~~
```
However, running it directly on my laptop does work without that header – as one of the other headers does include stdlib.h via some lengthy detour. Hence, I do not know how common this issue is.

> I think we should include the header as `<cstdlib>` rather than `<stdlib.h>` though, which looks to be better practice.

Done so now.

tob2:
> > I think we should include the header as `<cstdlib>` rather than `<stdlib.h>` though, which looks to be better practice.
> Done so now.

While this one works locally as well, it does fail on the tester with an internal compiler error :-( — I tried changing the order, but I now see that it fails for src/DepGraphNode.cpp.o and src/AlignedRangeIndexSetBuilders.cpp.o, which looks unrelated.

tom91136:
@tob2  There was an unrelated issue(#97) that broke the CI , can you merge `main` into `tob2:stdlib_fix` and see if it works?

tob2:
@tom91136 - rebased + pushed, now CI succeeds.

tob2:
@tomdeakin / @tom91136 - ping

tomdeakin:
Thanks for the reminder @tob2 - merged. Thanks for reporting and fixing this.

PR #102, opened 2021-05-23
Move all C++ impl. to ./cpp and remove Makefiles

tom91136:
This PR moves all C++ source files to the `./cpp` directory.
The CI and documentation has been updated to reflect this.

Motivation:
 * GitHub's README rendering is now very far down as the file list grows with new implementations. 
 * We might be adding new implementations in different programming languages (see #78, #95), having implementations grouped by language at the top-level helps with organisation.


andreipoe:
How about `src` instead of `cpp`?

tomdeakin:
This makes sense. I'd prefer `src/` too.

I wonder if we want to extend this and have `src/openmp`, `src/sycl`, etc...

tom91136:
So with `src`, we get something like:
```

**Comment by tomdeakin** [comment]
Lets go with that structure as a first pass.

What do we think about having each model in a subdirectory inside `src/`?

**Comment by andreipoe** [comment]
That may be a good idea. If we then have all the Makefile use standard names, we could compile with just `make -C OpenMP`.

**Comment by tom91136** [comment]
Yep, I'm doing that now, so we'll end up with a `Makefile` in each `/src/<impl>/`

**Comment by tom91136** [comment]
Ready for review, I've renamed the `<model>.cmake` files to just `model.cmake` so as to not duplicate the same information.

File: .gitignore [hunk 1]
```diff
@@ -1,30 +1,8 @@
...
-*.o
-*.bc
-*.sycl
-*.tar
-*.gz
-*.a
-
-KokkosCore_config.*
-
-.DS_Store
```
tomdeakin, line 21: Top level .gitignore should include this `.DS_Store`

.gitignore
```diff
@@ -1,30 +1,8 @@
...
-
-Makefile
 
 build/
 cmake-build-*/
 CMakeFiles/
 .idea/
 .vscode/
-.directory
```
tomdeakin, line 30: What is this for?

legacy/Makefile
```diff
@@ -14,7 +14,8 @@ endif
 
 
 hc-stream: ../main.cpp HCStream.cpp
```
tomdeakin, line 16: This file rename looks wrong. There might be multiple implementations placed in `legacy/` so we need to keep the Makefile name.

src/ci-test-compile.sh
```diff
@@ -86,38 +86,38 @@ run_build() {
 }
```
tomdeakin, line 86: The CI stuff isn't part of the BabelStream source - can this live in the top level directory?

src/ocl/Makefile
```diff
@@ -30,8 +30,8 @@ else
   LIBS = -lOpenCL
 endif
 
-ocl-stream: main.cpp OCLStream.cpp
-	$(CXX) $(CXXFLAGS) -DOCL $^ $(EXTRA_FLAGS) $(LIBS) -o $@
+ocl-stream: ../main.cpp OCLStream.cpp
+	$(CXX) $(CXXFLAGS) -DOCL $^ $(EXTRA_FLAGS) $(LIBS) -o $@ -I. -I..
```
tomdeakin, line 34: The local copy of the OpenCL headers in `CL` have been put in `src/CL` but they should go in `src/ocl/CL` instead.

src/omp/Makefile
```diff
@@ -91,8 +91,8 @@ endif
 
 OMP = $(OMP_$(COMPILER)_$(TARGET))
 
-omp-stream: main.cpp OMPStream.cpp
-	$(CXX) $(CXXFLAGS) -DOMP $^ $(OMP) $(EXTRA_FLAGS) -o $@
+omp-stream: ../main.cpp OMPStream.cpp
+	$(CXX) $(CXXFLAGS) -DOMP  $^ $(OMP) $(EXTRA_FLAGS) -o $@ -I. -I..
```
tomdeakin, line 95: Do we need to specify `-I.`? This comment applies to all the Makefiles where appropriate.

src/std/Makefile
```diff
@@ -6,8 +6,8 @@
 CXXFLAGS=-O3 -std=c++17 -stdpar -DSTD
 STD_CXX=nvc++
 
-std-stream: main.cpp STDStream.cpp
-	$(STD_CXX) $(CXXFLAGS) $^ $(EXTRA_FLAGS) -o $@
+std-stream: ../main.cpp STDStream.cpp
+	$(STD_CXX) $(CXXFLAGS) $^ $(EXTRA_FLAGS) -o $@ -I. -I..
```
tomdeakin, line 10: `-I.` shouldn't be required here as we're not including any local headers that aren't already included with `#include "name.h"`

src/std20/Makefile
```diff
@@ -17,8 +17,8 @@ FLAGS_GNU = -O3 -std=c++2a -march=native
 CXXFLAGS = $(FLAGS_$(COMPILER))
 
 
-std20-stream: main.cpp STD20Stream.cpp
-	$(CXX) -DSTD20 $(CXXFLAGS) $^ $(EXTRA_FLAGS) -o $@
+std20-stream: ../main.cpp STD20Stream.cpp
+	$(CXX) -DSTD20 $(CXXFLAGS) $^ $(EXTRA_FLAGS) -o $@ -I. -I..
```
tomdeakin, line 21: No need to `-I.`

src/sycl/Makefile
```diff
@@ -73,8 +73,8 @@ SYCL_LINK_FLAGS = $(SYCL_$(COMPILER)_LINK_FLAGS)
 SYCL_INCLUDE = $(SYCL_$(COMPILER)_INCLUDE)
 
 # only ComputeCpp generates .sycl files which is a bit odd to deal with so we opted to compile everything together
-sycl-stream: main.cpp SYCLStream.cpp
-	$(SYCL_SYCLCXX) $(SYCL_SYCLFLAGS) $(SYCL_FLAGS) $(SYCL_INCLUDE) -DSYCL $(EXTRA_FLAGS) $(SYCL_LINK_FLAGS) $^ -o $@
+sycl-stream: ../main.cpp SYCLStream.cpp
+	$(SYCL_SYCLCXX) $(SYCL_SYCLFLAGS) $(SYCL_FLAGS) $(SYCL_INCLUDE) -DSYCL $(EXTRA_FLAGS) $(SYCL_LINK_FLAGS) $^ -o $@ -I. -I..
```
tomdeakin, line 77: No need to `-I.`

src/sycl/SYCLStream.h
```diff
@@ -11,6 +11,15 @@
...
 
+#include "CL/opencl.h"
+
+// XXX Intel's SYCL impl. needs CL_MEM_CHANNEL_INTEL which is provided in dpcpp's include dir
+// however, depending the system configuration, the system CL header sometimes takes precedence
+// we only really need this macro to refer to the extension so this is probably OK
+#ifndef CL_MEM_CHANNEL_INTEL
+#define CL_MEM_CHANNEL_INTEL            0x4213
+#endif
+
```
tomdeakin, line 22: Changes like this should be done in the build system, not in the source code.
Even then, only if they are strictly required. If the CMake build finds the wrong header, then please change it so it requires the correct header.
I do not want to include a define copied from a vendor header in this repo!

src/sycl/model.cmake
```diff
@@ -48,7 +48,7 @@ macro(setup)
         set(ComputeCpp_DIR ${SYCL_COMPILER_DIR})
 
         # don't point to the CL dir as the imports already have the CL prefix
-        set(OpenCL_INCLUDE_DIR "${CMAKE_SOURCE_DIR}/CL")
+        set(OpenCL_INCLUDE_DIR "${CMAKE_SOURCE_DIR}")
```
tomdeakin, line 51: Likely need to change this after removing the include in the source.

tomdeakin (changes requested):

tomdeakin:
As we're about to merge in Julia, Rust, etc. I wonder if those should go into a `standalone` directory. Thoughts?

.gitignore
tom91136: KDE's Dolphin, and maybe some other file managers leave that, similar to `.DS_store`

tom91136:
Yeah, I'm OK with that. 
If we'll only ever have a few standalones ones then I think they can live at the top-level but it looks like we might add a few more later.
Not sure if reaching PRK levels of language/framework coverage is a future milestone.

I think we should merge all the standalones PRs (Julia, Rust, and maybe Scala) first and then I can figure out how to make this PR work with the CI.

tomdeakin:
Will resolve #110 on merge

tomdeakin:
This PR will be updated to do the following:
- Move all implementations to `./src/<model>/{*.cpp, ...}`
- Move standalone implementations to e.g. `./src/Scala/...`
- Remove Makefile support in favour of CMake only
- Update the CI to work with new directory structure
- Download CL headers on demand where required, and remove otherwise. Retain `cl.hpp` in new `src/ocl/`. This will resolve #109.

tom91136:
@tomdeakin CI is green now, ready for merge

src/sycl/SYCLStream.h
```diff
@@ -11,6 +11,15 @@
 
 #include "Stream.h"
 
+#include "CL/opencl.h"
```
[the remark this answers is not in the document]
tom91136: Removed

src/sycl/SYCLStream.h
tom91136: Removed

PR #103, opened 2021-05-23
Initial Scala implementation

tom91136:
This PR adds a pure Scala 3 implementation of BabelStream benchmark.
Ideally, this should be merged after #102.

This implementation makes extensive use of new Scala 3 features:

* Whitespace sensitive indentations
* New control syntax (e.g `for xs do ...`)
* Real™️ inline
* First class syntax for typeclasses

The README file contains detailed build instructions and detailed project walk through.

---

Some basic benchmarking has been done for prosperity.

We use the following JVMs:

* Graal EE 21.1.0
* Graal CE 21.1.0
* AdoptOpenJDK 11.0.11

Graal is Oracle's JVM implementation with a Java-based JIT engine via the JVMCI interface. The motivation is that it's easier to implement JIT optimisations in a managed language. The project also includes many other features such as native image compilation and cross-language interop. Graal EE is the commercial variant of Graal CE, the only major difference is the additional proprietary HIR/LIR optimisation passes (official documentation suggests more aggressive vectorisation).

AdoptOpenJDK is built directly from the OpenJDK upstream repository. Unlike Graal, OpenJDK features a more traditional (battle tested) C++ based JIT: C1 and C2.

We use the following flags for all JVMs:

```
-XX:-UseOnStackReplacement 
-XX:-TieredCompilation 
-XX:ReservedCodeCacheSize=512m 
-XX:+UseCompressedOops 
-Xms10g -Xmx10g
```

Our baseline is the OpenMP version of BabelStream compiled with GCC 10.1.0
with `-O3 -march=native -fopenmp`.

Legend:

* 1T - `PlainStream.scala`
* SPC - `ParStream.scala`
* Thread Pool - `ThreadedStream.scala`
* Java 8 Streams - `J8SStream.scala`

### Single socket results

1S grouped by kernel:
1S grouped by kernel
1S grouped by implementation/JVM:
1S grouped by implementation/JVM

As expected, Graal EE outperformed Graal CE and OpenJDK significantly when the CFG is simple, which is the case for the *1T* implementation, and by extension, the *Thread Pool* implementation as well.

***

For highly declarative abstraction like *Java 8 Streams* and *SPC* implementations, all JVMs show very similar performance. The achieved performance is relatively low (~50% of OpenMP), highlighting difficulty for the JVM to optimise away the added abstraction tax.

A quick inspection of what Graal EE/CE did with the kernels in the SPC implementation yielded the following:
The optimisation phases was unable to work out that the kernel body should be inlined as the actual loop is buried in several layers of `invokevirtual`. Graal then went on to optimise the kernel body to the extent possible without hoisting anything out. The effect of this is every iteration of loop will execute a long list of pointer decompression (if Compressed OOPS is enabled), bound checks on every array involved, memory barrier operations, and finally the kernel itself; the disassembly from Capstone shows a series of `cmp`/`je` to dynamic deoptimise addresses (>= 20 instructions) before actually executing the (sometimes) vectorised kernel (usually <= 5 instructions).

Surprisingly, the Dot kernel (the only kernel with a non-trivial reduction) achieved comparable performance to the OpenMP version is several JVMs. Further investigation suggests the measurement methodology, as ported from the C++ version, isn't suitable for use on a JVM. As we change the kernel order, the optimisation is applied very differently. A future version of the benchmark should investigate how [JMH](https://openjdk.java.net/projects/code-tools/jmh/) can be applied here while
keeping the required output format.

Sadly, even the fastest Scala implementation with a hand-rolled threading (i.e. worse productivity) was not able to match OpenMP.

Graal Native Image was also tested, but the performance was abnormally slow (maxing out at hundreds of MBs), so it is disqualified for now.

### Dual socket results

Finally, the same benchmark is ran on a 2S system to observe NUMA effects.

Picture3

**Observations**

Performance is poor compared to 1S, selecting NUMA-aware GCs (e.g. G1GC) did not make much of adifference.

**Future work**

This PR does not include any accelerator implementations. Potentially interesting library/frameworks
to look at:

* [Aparapi](https://aparapi.com/) - Userspace library
    * Bytecode via instrumentation to OpenCL source
* [TornadoVM](https://github.com/beehive-lab/TornadoVM) - Alternative JVM
    * Custom Graal passes via JVMCI with JIT to PTX/OpenCL

These will be added in follow-up PRs.


thorbenlouw:
Adding my comments here. Firstly, it's very nice work! 

Here are a few changes I think should go in:
(1) In the README, mention that you're using OpenHFT for thread affinity
(2) Also mention that you need to have a scala SDK, sbt etc installed (you only mention JDK and Graal). I think it's important since that's not a given on HPC systems people might be running this on, and building could be a PITA
(3) Short option '-n' is used for both numtimes and "nstream only" and your opt parser is fine with this
(4) AllKernels doesn't run nstream. Is that intentional?


Next up, some of the more controversial comments :
(a) I don't think you can make a claim that "_in theory, this implementation also covers Java_"  and then only use the Java 8 stddlib parallel streams. More like "we also cover the parallel IntStreams in the Java standard library". But you aren't testing here what javac might be doing to your code, and you haven't expressed it in Java. Remember part of the point of BabelStream is to test how compilers do on idiomatic descriptions of BabelStream in a programming paradigm. So with that in mind...
(b) I think you should (also? instead?) have a JavaStream that does the IntStreams, for comparison, written in Java. It would also be really good to get a "writing C in Java" (classic fors with mutation on primitives). Yes,  a Javastream will be faff, and potentially means other build systems and sdks. 

(c) It's totally nasty and super gross that you have to reinvent Fractional and Numeric  and devote so many LOC to obvious repetitions of operator definition, but it also looks unavoidable. Can you remind me why it's a problem to use 2.13's stdlib Fractional and Numeric?
BTW You'd have the same problem in a Java implementation. Both Double and Float inherit from Number, so you could have a interface BabelStream<T extends Number> ,but Number doesn't have * and + operators (only the autoboxed primitives do), so you'd have to implement overrides of some gross multiply  and sum methods and trust hotspot to inline here. Which it almost certainly will after some iterations I would guess.
In Java, I would almost go as far as living with a certain amount of expressed redundancy in the code to avoid this -- the abstraction costs almost as much code as a clearer duplication would, but this is opinion.
But this really ought not to be such a problem in Scala. Maybe as a compromise I'd suggest only using the parts of Fractional  that you need: a custom trait with just + and * ? YAGNI

(d) Collecting all the timings in memory before writing them out is probably fine - people are unlikely to be doing zillions of repeats, but since you never write out individual timings it makes more sense to be collecting the total runtime rather than holding all the individual times in memory right until the end

Questions:
One last thing (question): What happens if you oversaturate the number of workers in the Executor thread pool (more than one thread per core)? In theory it should be worse, right? Is it?

Most JVM benchmarks allow for a significant "burn-in" so that code which is likely going to be jitted and optimised, is already done by the time the timing starts. That's the only fair comparison between JVM and native. Did you consider adding an option for warm-up/burn-in, maybe default 20 iters or so?

Issue #74, opened 2020-07-15
Add TBB version


tomdeakin:

tomdeakin:
TBB implementation added in #105 .

PR #105, opened 2021-05-27
Initial TBB implementation

tom91136:
This PR adds the TBB (or oneTBB) implementation, thus closing #74.

The implementation is fairly straightforward as TBB exposes a very similar API to SYCL but without device/host memory distinctions.
We expose the *partitioner* parameter, common to most parallel algorithms, as the device option:

```shell
> tbb-stream --list
[0] auto partitioner
[1] affinity partitioner
[2] static partitioner
[3] simple partitioner
See https://spec.oneapi.com/versions/latest/elements/oneTBB/source/algorithms.html#partitioners for more details
```

Partitioners isn't an abstract class so TBB implements it as overloads for each exported `parallel_*` function.
To allow selecting partitioners as a runtime option, C++14 auto lambda parameters with forwarding was used to specialise for each case with minimal (very likely zero) runtime penalty.

Performance should be comparable to GCC's parallel STL implementation which calls TBB internally.
Benchmark results and scaling are coming soon, don't merge yet unless those aren't needed.

### DIFF
--- a/TBB.cmake
+++ b/TBB.cmake
@@ -1,8 +1,16 @@
 
-register_flag_required(TBB_DIR
-        "Absolute path to oneTBB distribution, the directory should contains at least `include/` and `lib/")
+register_flag_optional(ONE_TBB_DIR
+        "Absolute path to oneTBB (with header `onetbb/tbb.h`) distribution, the directory should contain at least `include/` and `lib/.
+         If unspecified, the system TBB (with header `tbb/tbb.h`) will be used via CMake's find_package(TBB)." 
+        "")
 
 macro(setup)
+    if(ONE_TBB_DIR)
+        set(TBB_ROOT "${ONE_TBB_DIR}") # see https://github.com/Kitware/VTK/blob/0a31a9a3c1531ae238ac96a372fec4be42282863/CMake/FindTBB.cmake#L34
+        # docs on Intel's website refers to TBB_DIR which hasn't been correct for 6 years
+    endif()
+    
+
     set(CMAKE_CXX_STANDARD 14) # we use auto in lambda parameters for the different partitioners
     # see https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md#tbbconfig---integration-of-binary-packages
     find_package(TBB REQUIRED)
--- a/TBB.cmake
+++ b/TBB.cmake
@@ -1,0 +1,10 @@
+
+register_flag_required(TBB_DIR
+        "Absolute path to oneTBB distribution, the directory should contains at least `include/` and `lib/")
+
+macro(setup)
+    set(CMAKE_CXX_STANDARD 14) # we use auto in lambda parameters for the different partitioners
+    # see https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md#tbbconfig---integration-of-binary-packages
+    find_package(TBB REQUIRED)
+    register_link_library(TBB::tbb)
+endmacro()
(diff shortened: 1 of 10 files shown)


TBBStream.hpp
```diff
@@ -0,0 +1,56 @@
...
+    std::vector<T> a;
+    std::vector<T> b;
+    std::vector<T> c;
+    
+
+    template < typename U, typename F>
+    U with_partitioner(const F &f);
+ 
+    template <typename F>
+    void parallel_for(const F &f);
```
tomdeakin, line 36: Can you manually inline the `parrallel_for` and `parallel_reduce` and remove these? We don't want to add our own abstraction on top.

TBBStream.hpp
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+#include <iostream>
+#include <vector>
+#include "tbb/tbb.h"
+#include "Stream.h"
+
+#define IMPLEMENTATION_STRING "TBB"
+
+enum class Partitioner : int { Auto = 0, Affinity, Static, Simple};
```
tomdeakin, line 16: These are an abstraction over the standard TBB ones. Can we just use one of those directly?

TBBStream.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+  std::cout << "Using TBB partitioner: " << getDeviceName(device) << std::endl;
+}
+
+template <class T>
+template <typename U, typename F>
+U TBBStream<T>::with_partitioner(const F &f) 
+{
+  switch(partitioner){
+    case Partitioner::Auto:      return f(tbb::auto_partitioner{});
```
tomdeakin, line 21: As elsewhere, if we store the TBB partitioner, these can all be invoked the same inline.

TBB.make
```diff
@@ -0,0 +1,28 @@
+
+ifndef COMPILER
+define compiler_help
+Set COMPILER to change flags (defaulting to GNU).
+Available compilers are:
+  GNU
```
tomdeakin, line 6: Can we add support for the Intel compiler too?

TBB.cmake
```diff
@@ -0,0 +1,18 @@
+
+register_flag_optional(ONE_TBB_DIR
+        "Absolute path to oneTBB (with header `onetbb/tbb.h`) distribution, the directory should contain at least `include/` and `lib/.
+         If unspecified, the system TBB (with header `tbb/tbb.h`) will be used via CMake's find_package(TBB)." 
+        "")
+
+macro(setup)
+    if(ONE_TBB_DIR)
+        set(TBB_ROOT "${ONE_TBB_DIR}") # see https://github.com/Kitware/VTK/blob/0a31a9a3c1531ae238ac96a372fec4be42282863/CMake/FindTBB.cmake#L34
+        # docs on Intel's website refers to TBB_DIR which hasn't been correct for 6 years
```
tomdeakin, line 10:
```suggestion
        # docs on Intel's website refers to TBB_DIR which is not correct
```

tomdeakin (changes requested):

tomdeakin:
Instead of exposing the partitioner as a runtime option and introducing a layer of abstraction and indirection, maybe it needs to be exposed as a compile time option in a similar way to how we choose the different CUDA allocation options.

TBBStream.hpp
tom91136: Fixed

TBBStream.hpp
tom91136: Fixed

TBBStream.cpp
tom91136: Fixed

TBB.make
tom91136: Added INTEL/INTEL_LEGACY, another PR for adding this to all other models coming soon

tom91136:
Ready for review again

tomdeakin (approved):
LGTM! Thanks @tom91136.

PR #106, opened 2021-06-10
Julia implementation

tom91136:
This PR adds the Julia implementation of BabelStream with the following implementations:

 * `PlainStream.jl` - Single threaded `for`
 * `ThreadedStream.jl` - Threaded implementation with `Threads.@threads` macros
 * `DistributedStream.jl` - Process based parallelism with `@distributed` macros
 * `CUDAStream.jl` - Direct port of BabelStream's native CUDA implementation using [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl)
 * `AMDGPUStream.jl` - Direct port of BabelStream's native HIP implementation using [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl)

See README.md for details on build and run instructions.

Performance is surprisingly good across all supported hardware platforms.
All benchmarks uses Julia 1.6.1, specific versions of each package are available in `Manifest.toml`

omp
cuda

AMDGPU.jl is currently still in heavy development, although the project reports most core features are working.
At the time of this PR, there are still a few issues that makes it unsuitable for production use:
 * Kernel performance is inconsistent (tested on on Radeon VII w/ ROCm 3.10)
 * API for device selection is missing, the implementation had to resort to introspection
 * Rapid kernel submission seems to cause ref-count overflow issues if we attempt to wait for each kernel to complete, this is a bit counter-intuitive. The workaround is to use hardware-based events for synchronisation.

hip
 

Finally, there isn't a process-based (e.g MPI) implementation of BabelStream so comparison for `DistributedStream.jl` has been omitted.
That said, performance seems to be significantly worst than `ThreadedStream.jl` due to the added serialisation overhead.

## Future work

We should be able to include [oneAPi.jl](https://github.com/JuliaGPU/oneAPI.jl) once it is ready for general use.

There's also [OpenCL.jl](https://github.com/JuliaGPU/OpenCL.jl) but it simply wraps the OpenCL host API; kernels must still be written in OpenCL C, so this wouldn't be any different from BabelStream's OCLStream.

### DIFF
--- a/JuliaStream.jl/README.md
+++ b/JuliaStream.jl/README.md
@@ -13,7 +13,7 @@ This is an implementation of BabelStream in Julia which contains the following v
 
 Prerequisites
 
- * Julia 1.6+
+ * Julia >= 1.6+
 
 With Julia on path, run the benchmark with:
 
@@ -24,7 +24,7 @@ With Julia on path, run the benchmark with:
 ```
 
 **Important:**
- * Julia is 1-indexed, so N > 1 in `--device N`
- * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1
- * `DistributedStream` uses `addprocs()` call directly which defaults to `$(nproc)`, **do not use the `-p <N>` flag** as per the [documentation](https://docs.julialang.org/en/v1/manual/distributed-computing).
+ * Julia is 1-indexed, so N >= 1 in `--device N`.
+ * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1.
+ * Worker count for `DistributedStream` is set with `-p <N>` as per the [documentation](https://docs.julialang.org/en/v1/manual/distributed-computing).
  * Certain implementations such as CUDA and AMDGPU will do hardware detection at runtime and may download and/or compile further software packages for the platform.
--- a/JuliaStream.jl/README.md
+++ b/JuliaStream.jl/README.md
@@ -1,0 +1,30 @@
+JuliaStream.jl
+==============
+
+This is an implementation of BabelStream in Julia which contains the following variants:
+
+ * `PlainStream.jl` - Single threaded `for`
+ * `ThreadedStream.jl` - Threaded implementation with `Threads.@threads` macros
+ * `DistributedStream.jl` - Process based parallelism with `@distributed` macros
+ * `CUDAStream.jl` - Direct port of BabelStream's native CUDA implementation using [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl)
+ * `AMDGPUStream.jl` - Direct port of BabelStream's native HIP implementation using [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl)
+
+### Build & Run
+
+Prerequisites
+
+ * Julia 1.6+
+
+With Julia on path, run the benchmark with:
+
+```shell
+> cd JuliaStream.jl
+> julia --project -e 'import Pkg; Pkg.instantiate()' # only required on first run
+> julia --project src/<IMPL>Stream.jl
+```
+
+**Important:**
+ * Julia is 1-indexed, so N > 1 in `--device N`
+ * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1
+ * You must *prepend* the number of processes needed for `DistributedStream`, e.g `julia -p$(nproc) --project src/DistributedStream.jl`
+ * Certain implementations such as CUDA and AMDGPU will do hardware detection at runtime and may download and/or compile further software packages for the platform.
--- a/JuliaStream.jl/README.md
+++ b/JuliaStream.jl/README.md
@@ -26,5 +26,5 @@ With Julia on path, run the benchmark with:
 **Important:**
  * Julia is 1-indexed, so N > 1 in `--device N`
  * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1
- * You must *prepend* the number of processes needed for `DistributedStream`, e.g `julia -p$(nproc) --project src/DistributedStream.jl`
+ * `DistributedStream` uses `addprocs()` call directly which defaults to `$(nproc)`, **do not use the `-p <N>` flag** as per the [documentation](https://docs.julialang.org/en/v1/manual/distributed-computing).
  * Certain implementations such as CUDA and AMDGPU will do hardware detection at runtime and may download and/or compile further software packages for the platform.
--- a/JuliaStream.jl/README.md
+++ b/JuliaStream.jl/README.md
@@ -8,6 +8,8 @@ This is an implementation of BabelStream in Julia which contains the following v
  * `DistributedStream.jl` - Process based parallelism with `@distributed` macros
  * `CUDAStream.jl` - Direct port of BabelStream's native CUDA implementation using [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl)
  * `AMDGPUStream.jl` - Direct port of BabelStream's native HIP implementation using [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl)
+ * `oneAPIStream.jl` - Direct port of BabelStream's native SYCL implementation using [oneAPI.jl](https://github.com/JuliaGPU/oneAPI.jl)
+ * `KernelAbstractions.jl` - Direct port of miniBUDE's native CUDA implementation using [KernelAbstractions.jl](https://github.com/JuliaGPU/KernelAbstractions.jl)
 
 ### Build & Run
 
@@ -15,12 +17,35 @@ Prerequisites
 
  * Julia >= 1.6+
 
-With Julia on path, run the benchmark with:
+A set of reduced dependency projects are available for the following backend and implementations:
+
+ * `AMDGPU` supports:
+   - `AMDGPUStream.jl`
+ * `CUDA` supports:
+   - `CUDAStream.jl`
+ * `oneAPI` supports:
+   - `oneAPIStream.jl`
+ * `KernelAbstractions` supports:
+   - `KernelAbstractionsStream.jl`
+ * `Threaded` supports:
+   - `PlainStream.jl`
+   - `ThreadedStream.jl`
+   - `DistributedStream.jl`
+
+With Julia on path, run your selected benchmark with:
 
 ```shell
 > cd JuliaStream.jl
-> julia --project -e 'import Pkg; Pkg.instantiate()' # only required on first run
-> julia --project src/<IMPL>Stream.jl
+> julia --project=<BACKEND> -e 'import Pkg; Pkg.instantiate()' # only required on first run
+> julia --project=<BACKEND> src/<IMPL>Stream.jl
+```
+
+For example. to run the CUDA implementation:
+
+```shell
+> cd JuliaStream.jl
+> julia --project=CUDA -e 'import Pkg; Pkg.instantiate()' 
+> julia --project=CUDA src/CUDAStream.jl
 ```
 
 **Important:**
@@ -28,3 +53,15 @@ With Julia on path, run the benchmark with:
  * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1.
  * Worker count for `DistributedStream` is set with `-p <N>` as per the [documentation](https://docs.julialang.org/en/v1/manual/distributed-computing).
  * Certain implementations such as CUDA and AMDGPU will do hardware detection at runtime and may download and/or compile further software packages for the platform.
+
+***
+
+Alternatively, the top-level project `Project.toml` contains all dependencies needed to run all implementations in `src`.
+There may be instances where some packages are locked to an older version because of transitive dependency requirements.
+
+To run the benchmark using the top-level project, run the benchmark with:
+```shell
+> cd JuliaStream.jl
+> julia --project -e 'import Pkg; Pkg.instantiate()'  
+> julia --project src/<IMPL>Stream.jl
+```
--- a/JuliaStream.jl/src/AMDGPUStream.jl
+++ b/JuliaStream.jl/src/AMDGPUStream.jl
@@ -6,32 +6,23 @@ const ROCData = StreamData{T,ROCArray{T}} where {T}
 const TBSize = 1024::Int
 const DotBlocks = 256::Int
 
-# AMDGPU.agents()'s internal iteration order isn't stable
-function gpu_agents_in_repr_order()
-  # XXX if we select anything other than :gpu, we get 
-  # HSA_STATUS_ERROR_INVALID_AGENT on the first kernel submission
-  sort(AMDGPU.get_agents(:gpu), by = repr)
-end
-
-function devices()
+function devices()::Vector{DeviceWithRepr}
   try
-    map(repr, gpu_agents_in_repr_order())
+    # AMDGPU.agents()'s internal iteration order isn't stable
+    sorted = sort(AMDGPU.get_agents(:gpu), by = repr)
+    map(x -> (x, repr(x), "AMDGPU.jl"), sorted)
   catch
     # probably unsupported
-    []
+    String[]
   end
-end
-
-function gridsize(data::ROCData{T})::Int where {T}
-  return data.size
 end
 
 function make_stream(
   arraysize::Int,
   scalar::T,
-  device::Int,
+  device::DeviceWithRepr,
   silent::Bool,
-)::ROCData{T} where {T}
+)::Tuple{ROCData{T},Nothing} where {T}
 
   if arraysize % TBSize != 0
     error("arraysize ($(arraysize)) must be divisible by $(TBSize)!")
@@ -39,30 +30,31 @@ function make_stream(
 
   # XXX AMDGPU doesn't expose an API for setting the default like CUDA.device!()
   # but AMDGPU.get_default_agent returns DEFAULT_AGENT so we can do it by hand
-  AMDGPU.DEFAULT_AGENT[] = gpu_agents_in_repr_order()[device]
-
-  data = ROCData{T}(
-    ROCArray{T}(undef, arraysize),
-    ROCArray{T}(undef, arraysize),
-    ROCArray{T}(undef, arraysize),
-    scalar,
-    arraysize,
-  )
+  AMDGPU.DEFAULT_AGENT[] = device[1]
   selected = AMDGPU.get_default_agent()
   if !silent
     println("Using GPU HSA device: $(AMDGPU.get_name(selected)) ($(repr(selected)))")
-    println("Kernel parameters   : <<<$(gridsize(data)),$(TBSize)>>>")
+    println("Kernel parameters   : <<<$(arraysize),$(TBSize)>>>")
   end
-  return data
+  return (
+    ROCData{T}(
+      ROCArray{T}(undef, arraysize),
+      ROCArray{T}(undef, arraysize),
+      ROCArray{T}(undef, arraysize),
+      scalar,
+      arraysize,
+    ),
+    nothing,
+  )
 end
 
-function init_arrays!(data::ROCData{T}, init::Tuple{T,T,T}) where {T}
+function init_arrays!(data::ROCData{T}, _, init::Tuple{T,T,T}) where {T}
   AMDGPU.fill!(data.a, init[1])
   AMDGPU.fill!(data.b, init[2])
   AMDGPU.fill!(data.c, init[3])
 end
 
-function copy!(data::ROCData{T}) where {T}
+function copy!(data::ROCData{T}, _) where {T}
   function kernel(a, c)
     i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
     @inbounds c[i] = a[i]
@@ -70,11 +62,11 @@ function copy!(data::ROCData{T}) where {T}
   end
   AMDGPU.wait(
     soft = false, # soft wait causes HSA_REFCOUNT overflow issues
-    @roc groupsize = TBSize gridsize = gridsize(data) kernel(data.a, data.c)
+    @roc groupsize = TBSize gridsize = data.size kernel(data.a, data.c)
   )
 end
 
-function mul!(data::ROCData{T}) where {T}
+function mul!(data::ROCData{T}, _) where {T}
   function kernel(b, c, scalar)
     i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
     @inbounds b[i] = scalar * c[i]
@@ -82,11 +74,11 @@ function mul!(data::ROCData{T}) where {T}
   end
   AMDGPU.wait(
     soft = false, # soft wait causes HSA_REFCOUNT overflow issues
-    @roc groupsize = TBSize gridsize = gridsize(data) kernel(data.b, data.c, data.scalar)
+    @roc groupsize = TBSize gridsize = data.size kernel(data.b, data.c, data.scalar)
   )
 end
 
-function add!(data::ROCData{T}) where {T}
+function add!(data::ROCData{T}, _) where {T}
   function kernel(a, b, c)
     i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
     @inbounds c[i] = a[i] + b[i]
@@ -94,11 +86,11 @@ function add!(data::ROCData{T}) where {T}
   end
   AMDGPU.wait(
     soft = false, # soft wait causes HSA_REFCOUNT overflow issues
-    @roc groupsize = TBSize gridsize = gridsize(data) kernel(data.a, data.b, data.c)
+    @roc groupsize = TBSize gridsize = data.size kernel(data.a, data.b, data.c)
   )
 end
 
-function triad!(data::ROCData{T}) where {T}
+function triad!(data::ROCData{T}, _) where {T}
   function kernel(a, b, c, scalar)
     i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
     @inbounds a[i] = b[i] + (scalar * c[i])
@@ -106,7 +98,7 @@ function triad!(data::ROCData{T}) where {T}
   end
   AMDGPU.wait(
     soft = false, # soft wait causes HSA_REFCOUNT overflow issues
-    @roc groupsize = TBSize gridsize = gridsize(data) kernel(
+    @roc groupsize = TBSize gridsize = data.size kernel(
       data.a,
       data.b,
       data.c,
@@ -115,7 +107,7 @@ function triad!(data::ROCData{T}) where {T}
   )
 end
 
-function nstream!(data::ROCData{T}) where {T}
+function nstream!(data::ROCData{T}, _) where {T}
   function kernel(a, b, c, scalar)
     i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
     @inbounds a[i] += b[i] + scalar * c[i]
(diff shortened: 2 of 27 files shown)


JuliaStream.jl/README.md
```diff
@@ -0,0 +1,30 @@
...
+```shell
+> cd JuliaStream.jl
+> julia --project -e 'import Pkg; Pkg.instantiate()' # only required on first run
+> julia --project src/<IMPL>Stream.jl
+```
+
+**Important:**
+ * Julia is 1-indexed, so N > 1 in `--device N`
+ * Thread count for `ThreadedStream` must be set via the `JULIA_NUM_THREADS` environment variable (e.g `export JULIA_NUM_THREADS=$(nproc)`) otherwise it defaults to 1
+ * `DistributedStream` uses `addprocs()` call directly which defaults to `$(nproc)`, **do not use the `-p <N>` flag** as per the [documentation](https://docs.julialang.org/en/v1/manual/distributed-computing).
```
tomdeakin, line 29: What is the reason behind this? Why don't we want to follow the documented approach of using the `-p` flag?

JuliaStream.jl/src/AMDGPUStream.jl
```diff
@@ -0,0 +1,178 @@
...
+  )
+  selected = AMDGPU.get_default_agent()
+  if !silent
+    println("Using GPU HSA device: $(AMDGPU.get_name(selected)) ($(repr(selected)))")
+    println("Kernel parameters   : <<<$(gridsize(data)),$(TBSize)>>>")
+  end
+  return data
+end
+
+function hard_wait(kernel)
```
tomdeakin, line 59: Can we inline this so we don't introduce extra functions that look like they are APIs?

JuliaStream.jl/src/AMDGPUStream.jl
```diff
@@ -0,0 +1,178 @@
...
+    i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x
+    @inbounds c[i] = a[i]
+    return
+  end
+  hard_wait(@roc groupsize = TBSize gridsize = gridsize(data) kernel(data.a, data.c))
+end
+
+function mul!(data::ROCData{T}) where {T}
+  function kernel(b, c, scalar)
+    i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x
```
tomdeakin, line 81: This looks weird. Why is the workgroup Idx -1 but not the workitem Idx? If workgroups are numbered from 1 (or 2) but work items are numbers from 0 (or 1), it's worth adding a comment.

JuliaStream.jl/src/AMDGPUStream.jl
```diff
@@ -0,0 +1,178 @@
...
+    )
+  )
+  return sum(partial_sum)
+end
+
+function read_data(data::ROCData{T})::VectorData{T} where {T}
+  return VectorData{T}(data.a, data.b, data.c, data.scalar, data.size)
+end
+
+main()
```
tomdeakin, line 178: Do we need `main()` here?

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,146 @@
...
+
+function init_arrays!(data::CuData{T}, init::Tuple{T,T,T}) where {T}
+  CUDA.fill!(data.a, init[1])
+  CUDA.fill!(data.b, init[2])
+  CUDA.fill!(data.c, init[3])
+end
+
+function copy!(data::CuData{T}) where {T}
+  function kernel(a, c)
+    i = (blockIdx().x - 1) * blockDim().x + threadIdx().x
```
tomdeakin, line 58: Same as with the AMDGPU one, worth checking the blocks and threads are indexed differently, adding a comment if necessary.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,146 @@
...
+const CuData = StreamData{T,CuArray{T}} where {T}
+const TBSize = 1024::Int
+const DotBlocks = 256::Int
+
+function devices()
+  return !CUDA.functional(false) ? [] :
+         map(d -> "$(CUDA.name(d)) ($(repr(d)))", CUDA.devices())
+end
+
+function blocks(data::CuData{T})::Int where {T}
```
tomdeakin, line 13: What does this do and why do we need it? Again, we probably want to inline this so we don't introduce something that looks like a new API call. Might need a comment if it's an extra step we need to be aware of.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,146 @@
...
+  @cuda blocks = DotBlocks threads = TBSize kernel(data.a, data.b, data.size, partial_sum)
+  CUDA.synchronize()
+  return sum(partial_sum)
+end
+
+function read_data(data::CuData{T})::VectorData{T} where {T}
+  return VectorData{T}(data.a, data.b, data.c, data.scalar, data.size)
+end
+
+main()
```
tomdeakin, line 146: Again, do we need `main()`?

JuliaStream.jl/src/DistributedStream.jl
```diff
@@ -0,0 +1,86 @@
...
+  return @distributed (+) for i = 1:data.size
+    @inbounds data.a[i] * data.b[i]
+  end
+end
+
+function read_data(data::SharedArrayData{T})::VectorData{T} where {T}
+  return VectorData{T}(data.a, data.b, data.c, data.scalar, data.size)
+end
+
+main()
```
tomdeakin, line 86: Do we need `main()`?

JuliaStream.jl/src/DistributedStream.jl
```diff
@@ -0,0 +1,86 @@
+using Distributed
+
+include("Stream.jl")
+
+addprocs()
```
tomdeakin, line 5: See comment in README

JuliaStream.jl/src/PlainStream.jl
```diff
@@ -0,0 +1,64 @@
...
+    @inbounds sum += data.a[i] * data.b[i]
+  end
+  return sum
+end
+
+function read_data(data::VectorData{T})::VectorData{T} where {T}
+  return data
+end
+
+main()
```
tomdeakin, line 64: Do we need `main()`?

JuliaStream.jl/src/Stream.jl
```diff
@@ -0,0 +1,292 @@
...
+    elseif benchmark == Triad
+      gold_a = gold_b + data.scalar * gold_c
+    elseif benchmark == Nstream
+      gold_a += gold_b + data.scalar * gold_c
+    else
+      error("Unknown benchmark", benchmark)
+    end
+  end
+
+  tolerance = eps(T) * 100
```
tomdeakin, line 71: Do we need to worry about making sure `100` is floating point here?

JuliaStream.jl/src/StreamData.jl
```diff
@@ -0,0 +1,8 @@
+
```
tomdeakin, line 1: Why do we need to wrap up the data arrays in a whole new struct?

tomdeakin (changes requested):
Interesting to see this - think it's pretty close. In general I'd like to make sure this is idiomatic of Julia code and we're not introducing any extra layers of abstraction in the benchmark.

JuliaStream.jl/src/AMDGPUStream.jl
tom91136: It's needed otherwise nothing runs when you load the script, `main()` isn't a special function the the runtime calls for you, see https://stackoverflow.com/a/35387566/896997

JuliaStream.jl/src/AMDGPUStream.jl
tom91136: See comment on the CUDA one. Will add a comment in code.

JuliaStream.jl/src/CUDAStream.jl
tom91136: Only blocks use 1-indexing, I'll add a comment in code. For reference, towards the end of the [section](https://cuda.juliagpu.org/stable/tutorials/introduction/#Writing-a-parallel-GPU-kernel), it used `index = (blockIdx().x - 1) * blockDim().x + threadIdx().x` for the global index.
This seems to be the preferred way of doing it, also seen in the [examples](https://github.com/JuliaGPU/CUDA.jl/blob/13dcce7415b7ed631208b00e901a6e6f76d11bf0/examples/vadd.jl#L6) and [internally](https://github.com/JuliaGPU/CUDA.jl/blob/13dcce7415b7ed631208b00e901a6e6f76d11bf0/src/mapreduce.jl#L144).

JuliaStream.jl/src/AMDGPUStream.jl
tom91136: PS: This implementation isn't a module and can't be used as one so we're not using the `__init__`

JuliaStream.jl/src/PlainStream.jl
tom91136: See comment above

JuliaStream.jl/src/DistributedStream.jl
tom91136: See comment above

JuliaStream.jl/src/Stream.jl
tom91136: I haven't used Julia long enough to know whether doing it like this (via multiple dispatch of the `*` function) is idiomatic. Since Julia's documentation mentions multiple dispatch as Julia's forte, I think this is probably an OK way.
It's also safe:
```
julia> typeof(eps(Float64) * 1)
Float64
``` 
The alternative would be to `convert` first:
```
julia> typeof(convert(Float64, 100))
Float64
julia> eps(Float64) * convert(Float64, 100)
2.220446049250313e-14
```
I'm leaning towards keeping `* 100`

JuliaStream.jl/src/StreamData.jl
tom91136: It's similar to what we have for Rust, there are no classes in Julia, only functions and data (structs).
The alternative is to pass everything as parameters:
```
function mul!(data::StreamData{T}) where {T}
```
becomes
```
function mul!(a::C, b::C, c::C, scalar::T, size::Int) where {T,C<:AbstractArray{T}}
```
Same applies when I copy everything back to host:
```
const ROCData = StreamData{T,ROCArray{T}} where {T}
...
function read_data(data::ROCData{T})::VectorData{T} where {T}
  return VectorData{T}(data.a, data.b, data.c, data.scalar, data.size)
end
```
`read_data` would be very long, and it needs to be implemented for 5 different implementations.

JuliaStream.jl/README.md
tom91136: Adding `-p` forks the process before the project environment gets initialised.
The alternative is to setup the environment for each worker process, this lengthens the already slow startup. 

That said, the distributed version is already quite slow so it's probably fine. 
It's less flexible with `addprocs()` anyway as I'll have to propagate the worker count myself from the program args which isn't implemented.

JuliaStream.jl/src/AMDGPUStream.jl
tom91136: Done

JuliaStream.jl/src/CUDAStream.jl
tom91136: It saves us from repeating `data.size ÷ TBSize` 6 times, it also needs the `÷` for integer division which is a pain to work with.
Given it's length, it does look to more readable when inlined.

JuliaStream.jl/README.md
tom91136: Done

JuliaStream.jl/src/DistributedStream.jl
tom91136: Done

JuliaStream.jl/src/CUDAStream.jl
tom91136: Done

JuliaStream.jl/src/AMDGPUStream.jl
tom91136: Done

tom91136:
Ready for review again.

JuliaStream.jl/src/StreamData.jl
tomdeakin: I think I prefer passing things as parameters.

tomdeakin:
Thanks @tom91136. I think I prefer the parameter passing rather than making a structure just to hold the arrays. I think that in a larger code with more arrays, you're just going to have to pass things around rather than keep wrapping things up in bundles to pass to different functions. Ideally we're aiming to write BabelStream in a way that is representative of something much bigger.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,152 @@
+include("Stream.jl")
+using CUDA
+
+const CuData = StreamData{T,CuArray{T}} where {T}
+const TBSize = 1024::Int
+const DotBlocks = 256::Int
+
+function devices()
+  return !CUDA.functional(false) ? [] :
```
giordano, line 9: Minor nitpick: make the function type-stable
```suggestion
  return !CUDA.functional(false) ? String[] :
```

JuliaStream.jl/src/AMDGPUStream.jl
```diff
@@ -0,0 +1,181 @@
...
+  # HSA_STATUS_ERROR_INVALID_AGENT on the first kernel submission
+  sort(AMDGPU.get_agents(:gpu), by = repr)
+end
+
+function devices()
+  try
+    map(repr, gpu_agents_in_repr_order())
+  catch
+    # probably unsupported
+    []
```
giordano, line 21:
```suggestion
    String[]
```

JuliaStream.jl/src/DistributedStream.jl
```diff
@@ -0,0 +1,85 @@
...
+function init_arrays!(data::SharedArrayData{T}, init::Tuple{T,T,T}) where {T}
+
+  @sync @distributed for i = 1:data.size
+    @inbounds data.a[i] = init[1]
+    @inbounds data.b[i] = init[2]
+    @inbounds data.c[i] = init[3]
+  end
+end
+
+function copy!(data::SharedArrayData{T}) where {T}
```
giordano, line 45: Note that here and in the functions below where `T` isn't really used to constrain the signature nor in the body of the function, the parameter is redundant and
```suggestion
function copy!(data::SharedArrayData)
```
would work just fine

JuliaStream.jl/src/Stream.jl
```diff
@@ -0,0 +1,291 @@
...
+    error(
+      "Device $(config.device) out of range (1..$(length(ds))), NOTE: Julia is 1-indexed",
+    )
+  end
+
+  if config.float
+    type = Float32
+  else
+    type = Float64
+  end
```
giordano, line 181: For the record, you can also do
```suggestion
  type = if config.float
    Float32
  else
    Float64
  end
```
or even
```suggestion
  type = config.float ? Float32 : Float64
```
if you want to be even more concise.  Also in the `if` below you can save a couple of `benchmark =` in the same way

giordano:
> Performance is surprisingly good across all supported hardware platforms.

:smiley: 

Is there something we can do to move this forward?  I had a _very_ quick look, but could do a more thorough review, if that helps

tom91136:
@giordano Thanks for the review! This PR will be used for an upcoming submission to [PMBS](http://www.pmbsworkshop.org/) so I got a few more local changes (I've added a functional [oneAPI.jl](https://github.com/JuliaGPU/oneAPI.jl) and [KA](https://github.com/JuliaGPU/KernelAbstractions.jl) implementation) that I'm in the process of finalising. I'll incorporate your review and put up a final version for further review by the end of the week.
If you're interested, the PMBS submission will also include a [compute bound benchmark](https://github.com/UoB-HPC/miniBUDE) written in Julia.

@tomdeakin and I had a discussion on the parameter passing and I think we've settled on the current approach being acceptable.

jpsamaroo (commented):
Cool stuff! Let me know if I can help with anything AMDGPU.jl-related (I'm the author and maintainer of it).

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,154 @@
...
+      CuArray{T}(undef, arraysize),
+      scalar,
+      arraysize,
+    ),
+    nothing,
+  )
+end
+
+function init_arrays!(data::CuData{T}, _, init::Tuple{T,T,T}) where {T}
+  CUDA.fill!(data.a, init[1])
```
maleadt, line 49: You shouldn't need the module prefix here, as `fill!` is just a Base method that dispatches correctly since you're passing a CuArray argument already. Only when that's not the case (e.g. with `fill`, which just takes an integer) you need the CUDA-specific version.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,154 @@
...
+  CUDA.fill!(data.b, init[2])
+  CUDA.fill!(data.c, init[3])
+end
+
+function copy!(data::CuData{T}, _) where {T}
+  function kernel(a, c)
+    i = (blockIdx().x - 1) * blockDim().x + threadIdx().x # only blockIdx starts at 1
```
maleadt, line 56: `threadIdx` also starts at 1, but our arrays are 1-indexed, so everything works out here.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,154 @@
...
+end
+
+function mul!(data::CuData{T}, _) where {T}
+  function kernel(b, c, scalar)
+    i = (blockIdx().x - 1) * blockDim().x + threadIdx().x # only blockIdx starts at 1
+    @inbounds b[i] = scalar * c[i]
+    return
+  end
+  @cuda blocks = data.size ÷ TBSize threads = TBSize kernel(data.b, data.c, data.scalar)
+  CUDA.synchronize()
```
maleadt, line 71: I'm not familiar with the underlying architecture here, but do you really need to synchronize after every kernel? That's not great for performance. Our GPU-to-CPU copy already synchronizes, so typically you don't need anything else.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,154 @@
...
+
+    if (local_i == 1)
+      @inbounds partial[blockIdx().x] = tb_sum[local_i]
+    end
+
+    return
+  end
+  partial_sum = CuArray{T}(undef, DotBlocks)
+  @cuda blocks = DotBlocks threads = TBSize kernel(data.a, data.b, data.size, partial_sum)
+  CUDA.synchronize()
```
maleadt, line 146: The sum below will run on the same stream as this kernel, so you don't need the synchronization.

JuliaStream.jl/src/oneAPIStream.jl
```diff
@@ -0,0 +1,170 @@
...
+  @oneapi items = groupsize groups = data.size ÷ groupsize kernel( #
+    data.b,
+    data.c,
+    data.scalar,
+  )
+  oneAPI.synchronize()
+end
+
+function add!(data::oneData{T}, groupsize::Int) where {T}
+  function kernel(a, b, c)
```
maleadt, line 83: It's a very minor detail, but you might want to put the `T` typevar on the inner kernel function instead, e.g., by using `a::AbstractArray{T}, ...` as arguments. That way you enforce specialization of the kernel method, whereas the outer CPU method doesn't necessarily need to be specialized (and you can use `eltype(data)` if you need the element type there).

But these things only matter when you work with very complicated inputs.

JuliaStream.jl/src/CUDAStream.jl
```diff
@@ -0,0 +1,154 @@
...
+  scalar::T,
+  device::DeviceWithRepr,
+  silent::Bool,
+)::Tuple{CuData{T},Nothing} where {T}
+
+  if arraysize % TBSize != 0
+    error("arraysize ($(arraysize)) must be divisible by $(TBSize)!")
+  end
+
+  # so CUDA's device is 0 indexed, so -1 from Julia
```
maleadt, line 24: CUDA.jl's `device!` is 0-indexed too.

maleadt (commented):
Looks good! I added a couple of small comments. Great to see that the performance is good :slightly_smiling_face:

JuliaStream.jl/src/CUDAStream.jl
tom91136: Oops, the comment was outdated as we switched to passing device instances instead of index around.

JuliaStream.jl/src/CUDAStream.jl
tom91136: The original implementation issues `cudaDeviceSynchronize();` after kernel submission (see https://github.com/UoB-HPC/BabelStream/blob/8f9ca7baa77c874897e1691c895729a39d959012/CUDAStream.cu#L161) so this port must match that. Does the `synchronize` in Julia does a lot more than `cudaDeviceSynchronize()`?  From what I can see, synchronisation doesn't seem to add any more overhead compare to the native version.

JuliaStream.jl/src/oneAPIStream.jl
tom91136: Does this apply to all other GPU implementations or just oneAPI? 
Ideally this would be an exemplar for writing kernels so would you recommend I annotate every other GPU implementation like this?
 
Aside:
I thought about doing this with more precise types (e.g `::CuDeviceVector{T}` for CUDA) but felt like the actual type is an implementation detail for the kernel launch macro  (`@cuda`, `@oneapi`, etc) so I guess `AbstractArray{T}` is the best we an do.

JuliaStream.jl/src/AMDGPUStream.jl
```diff
@@ -0,0 +1,181 @@
...
+end
+
+function copy!(data::ROCData{T}) where {T}
+  function kernel(a, c)
+    i = (workgroupIdx().x - 1) * workgroupDim().x + workitemIdx().x  # only workgroupIdx starts at 1
+    @inbounds c[i] = a[i]
+    return
+  end
+  AMDGPU.wait(
+    soft = false, # soft wait causes HSA_REFCOUNT overflow issues
```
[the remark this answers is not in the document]
tom91136: I've tested it with and without soft wait, and for the very few times `soft=true` actually run to completion (in most cases it dies after 70 iterations, so 5*70 kernels) the performance seems within margin of error. I think I'll stick to `soft=false` for now.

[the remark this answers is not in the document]
jpsamaroo: Can you explain what you mean by:
> actually run to completion (in most cases it dies after 70 iterations, so 5*70 kernels)

If you have a stacktrace or other information, that would help greatly.

JuliaStream.jl/src/oneAPIStream.jl
maleadt: This applies to all Julia code, not just the oneAPI GPU implementation. If you add a typevar (`::T .. .where T`) you force the compiler to specialize. That's needed for the GPU kernel, because we don't want or support dynamic code for GPUs, but the host function (here, `add!`) may remain underspecialized if the compiler thinks that's useful.

And yeah, the `AbstractArray` suggestion is so that you don't have to deal with the exact device-side types, which are an implementation detail.

JuliaStream.jl/src/CUDAStream.jl
maleadt: It should have similar overhead.

JuliaStream.jl/src/ThreadedStream.jl
```diff
@@ -63,12 +63,46 @@ function nstream!(data::VectorData{T}, _) where {T}
...
+    @fastmath for i = startidx:endidx
+      @inbounds acc += data.a[i] * data.b[i]
+    end
+    @inbounds partial[group] = acc
   end
   return sum(partial)
+  # This doesn't do well on aarch64 because of the excessive Threads.threadid() ccall 
+  # and inhibited vectorisation from the lack of @fastmath 
+  # partial = zeros(T, Threads.nthreads())
+  # Threads.@threads for i = 1:data.size
```
vchuravy, line 102: Normally I would do a:

```
@threads for tid in 1:nthreads()
    region = ....
    acc = 0.0
    @simd for i in region
       acc += ...
    end
    partial[i] = acc
end
```

Or using @tkf FLoops.jl/Folds

JuliaStream.jl/src/ThreadedStream.jl
```diff
@@ -63,12 +63,46 @@ function nstream!(data::VectorData{T}, _) where {T}
...
+end
+
 function dot(data::VectorData{T}, _) where {T}
-  partial = zeros(T, Threads.nthreads())
-  Threads.@threads for i = 1:data.size
-    @inbounds partial[Threads.threadid()] += data.a[i] * data.b[i]
+  partial = Vector{T}(undef,  Threads.nthreads())
+  static_par_ranged(data.size, Threads.nthreads()) do group, startidx, endidx
+    acc = zero(T)
+    @fastmath for i = startidx:endidx
```
vchuravy, line 93: Use `@simd` instead of `@fastmath` that makes sure that only the reduction variable get's reordered

JuliaStream.jl/src/ThreadedStream.jl
tom91136: Is that semantically equivalent to the `static_par_ranged`?  (minus the `@simd` bit)

JuliaStream.jl/src/ThreadedStream.jl
tom91136: Thanks! I didn't really want to use `@fastmath` in a benchmark like STREAM and I'm happy report I get the same runtime with `@simd` on both ThunderX2 and A64fx!

JuliaStream.jl/src/ThreadedStream.jl
vchuravy: I think so. Basically that's the canonical example of saying, I want one iteration per thread. Then you have to do the splitting manually, but you also don't have to reimplement the lower-level details xD

vchuravy: https://github.com/JuliaFolds/FLoops.jl#parallel-loop is an alternative

JuliaStream.jl/src/ThreadedStream.jl
giordano: On the A64FX node of Isambard, I've got a good speed up with `@simd` when setting the environment variable `JULIA_LLVM_ARGS="-aarch64-sve-vector-bits-min=512"`: https://github.com/JuliaLang/julia/issues/40308#issuecomment-901478623. Note that this requires llvm 12, so Julia v1.7

vchuravy: Yeah `@simd` finds the reduction chain, and then adds a subset of the fast math flags onto that chain.

For the A64fx you might also want to set `JULIA_LLVM_ARGS="-aarch64-sve-vector-bits-min=512"` see https://github.com/JuliaLang/julia/issues/40308#issuecomment-901478623

giordano: BTW, I didn't show in that comment that there is also a good scaling with the size of the floating type (including half-precision `Float16`):
```julia
$ JULIA_LLVM_ARGS="-aarch64-sve-vector-bits-min=512" julia -q
julia> using BenchmarkTools

julia> function sumsimd(x)
           s = zero(eltype(x))
           @simd for xi in x
               s += xi
           end
           s
       end
sumsimd (generic function with 1 method)

julia> @btime sumsimd(x) setup=(x = randn(Float64, 1_000_000))
  191.912 μs (0 allocations: 0 bytes)
1853.0335322487956

julia> @btime sumsimd(x) setup=(x = randn(Float32, 1_000_000))
  80.330 μs (0 allocations: 0 bytes)
400.9806f0

julia> @btime sumsimd(x) setup=(x = randn(Float16, 1_000_000))
  42.761 μs (0 allocations: 0 bytes)
Float16(1.872e3)
```

Related to this, you can watch this lightning talk at JuliaCon 2021 about speedup on using `Float16` on Isambard: https://www.youtube.com/watch?v=btHfZr2C0GA

vchuravy: Note that the speedup reported at JuliaCon requires https://github.com/JuliaLang/julia/issues/40216

JuliaStream.jl/src/ThreadedStream.jl
tkf: Yeah, it'd be better to use `@threads for i 1:nthreads()` than touching internals like `task.sticky = true` and `jl_set_task_tid`.

The main difference to FLoops.jl is that it lets the Julia runtime dynamic load-balancing. It doesn't matter if you only micro-benchmarking a parallel dot product but static scheduling like `@threads for` and `static_par_ranged` would fail when used inside other parallel functions (it has to wait for all worker threads a re available). So, _if_ the aim is to benchmark practical and composable pieces of routines in Julia, it may be better to use dynamic scheduling. On the other hand, if you want to benchmark what you can squeeze out from the Julia runtime at the moment, static scheduling like `@threads for` and `static_par_ranged` may be fine.

Another difference to FLoops.jl is that, by default, it uses divide-and-conquer (DAC) scheduling (`@spawn`s and `wait`s) while `@threads for` and `static_par_ranged`  use sequential scheduling. I find that DAC (in Julia) sometimes helps performance since it has larger parallelism (in the sense of work/span ratio), especially when there are a lot of cores.

PR #127, opened 2022-04-01
WIP: support massive input sizes

tomdeakin:
As main memory sizes increase, we are seeing errors for very large input sizes, passed in via the command line argument `--arraysize`. This reads in an `int` which can store values up to 2,147,483,647 (`2^31 - 1`).

First, the `main.cpp` driver code is updated to parse and store the input as a `intptr_t`. This gives us a range up to all of memory.

Secondly, the OpenMP implementation is updated to:
- Store the array size as a `intptr_t` inside the class (note implementations own their own data and array size).
- Update the loop bounds of all kernels to use `intptr_t` indexing. Note we want to keep using a signed data type because it makes vectorisation easier because the [loop is countable](https://www.ibm.com/docs/en/xl-c-and-cpp-aix/11.1.0?topic=programs-countable-loops) (e.g. [MSVC](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/vectorizer-and-parallelizer-messages?view=msvc-170#BKMK_ReasonCode100x), [Intel](https://community.intel.com/t5/Intel-C-Compiler/Vectorization-failed-because-of-unsigned-integer/td-p/1105071))

Finally, the starting literal values need the long double suffix: e.g. `0.2L`.

This PR currently updates only the OpenMP code. This change needs propagating to the other models.

## Progress

- [ ] OpenACC
- [ ] CUDA
- [ ] HIP
- [ ] Java ???
- [ ] Julia ???
- [ ] Kokkos
- [ ] OpenCL
- [X] OpenMP
- [ ] Raja
- [ ] Rust ???
- [ ] Scala
- [ ] std-data (C++ stdpar)
- [ ] std-indices (C++ stdpar)
- [ ] std-ranges (C++ stdpar)
- [ ] SYCL
- [ ] SYCL 2020
- [ ] TBB
- [ ] Thrust

### DIFF
--- a/src/Stream.h
+++ b/src/Stream.h
@@ -11,10 +11,10 @@
 #include <string>
 
 // Array values
-#define startA (0.1)
-#define startB (0.2)
-#define startC (0.0)
-#define startScalar (0.4)
+#define startA (0.1L)
+#define startB (0.2L)
+#define startC (0.0L)
+#define startScalar (0.4L)
 
 template <class T>
 class Stream
--- a/src/ci-prepare-bionic.sh
+++ b/src/ci-prepare-bionic.sh
@@ -133,21 +133,26 @@ setup_aocc() {
 }
 
 setup_nvhpc() {
-  echo "Preparing Nvidia HPC SDK"
-  local tarball="nvhpc.tar.gz"
-#  local url="http://localhost:8000/nvhpc_2021_219_Linux_x86_64_cuda_11.4.tar.gz"
-  local url="https://developer.download.nvidia.com/hpc-sdk/21.9/nvhpc_2021_219_Linux_x86_64_cuda_11.4.tar.gz"
+ echo "Preparing Nvidia HPC SDK"
+  local nvhpc_ver="22.3"
+  local nvhpc_release="2022_223"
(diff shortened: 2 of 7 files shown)


tom91136:
@tomdeakin I wonder whether we should template/macro the induction or not and make it a runtime option (e.g `--counter=uint32_t|uint64_t|intptr_t` ). It's quite possible changing the loop induction's type will alter how it's optimised (positively or negatively) so leaving it as an option allows us to compare with results from before this change.

tomdeakin:
We definitely don't want to offer unsigned types here. We need signed types to help with vectorisation, and don't want to suggest that using unsigned is best practice. Given that, we have limited choices of the type we should use.
* `intptr_t`
* `int64_t`
* `long long`

`intptr_t` guarantees we can hold a pointer to any place in memory, so the pointer arithmetic should always be valid.

tomdeakin:
Resolve this for v6.0 - I'm to 100% this is the right approach to solve this with `intptr_t`. We need to review other approaches to this (e.g., Kokkos, C++)

tomdeakin:
Closing as efforts are superseded by the approach in #188

PR #135, opened 2022-08-03
Fortran ports

jeffhammond:
This is a new implementation of BabelStream using Fortran.

The code uses a Fortran driver that is largely equivalent to the C++ one, with a few exceptions.  First, it does not use a C++ class for the `stream` object, since that doesn't seem like a useful way to do things in Fortran.  Instead, I use a module that contains the same methods, and which has `alloc` and `dealloc` that act like CTOR and DTOR.

The current implementations are:
* `DO CONCURRENT`
* Fortran array notation
* Sequential `DO` loops
* OpenACC `parallel loop`
* OpenACC `kernels` on Fortran array notation
* OpenMP `parallel do`
* OpenMP `taskloop`
* OpenMP `target teams distribute parallel do simd`
* OpenMP `target teams loop`
* CUDA Fortran (handwritten CUDA Fortran kernels, except DOT)
* CUDA Fortran kernels (`!$cuf kernel do <<<*,*>>>`)

I have tested with GCC, Intel (ifort and ifx), and NVHPC compilers on AArch64, x86_64 and NVIDIA GPU targets, although not exhaustively.

The current build system is GNU Make, and requires the user to manually specify the compiler and implementation.  I have not, and will not, do anything related to CMake.

~The only thing missing now is CSV output.~ CSV printing is supported.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,12 +10,18 @@ sycl-stream
 hip-stream
 tbb-stream
 
+src/fortran/BabelStream
+src/fortran/BabelStream.*
+
 *.o
 *.bc
 *.sycl
 *.tar
 *.gz
 *.a
+*.mod
+*.cub
+*.ptx
 
 KokkosCore_config.*
 
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -16,6 +16,8 @@
 **/*.gz
 **/*.a
 
+**/*.swp
+
 **/KokkosCore_Config_*
 
 **/.DS_Store
--- a/src/fortran/ArrayStream.F90
+++ b/src/fortran/ArrayStream.F90
@@ -1,0 +1,105 @@
+module ArrayStream
+    use, intrinsic :: ISO_Fortran_env
+    use BabelStreamTypes
+
+    implicit none
+
+    character(len=5), parameter :: implementation_name = "Array"
+
+    integer(kind=StreamIntKind) :: N
+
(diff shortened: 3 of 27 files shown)


jefflarkin:
Comment to the reviewers: For the array version, would you consider adding `!$acc kernels` around the sections of array syntax? In theory, it would also be valid to put an `!$omp workshare` around them for OpenMP compilers. I'm not sure whether any compilers will auto-parallelize, much less auto-offload, the array syntax, but a variety of compilers will happily support one or both of those parallelization hints.

tomdeakin:
@jefflarkin - definitely worth exploring what happens in both of those cases for sure!

jeffhammond:
@jefflarkin See OpenACCArrayStream.F90

jeffhammond:
OpenMPWorkshareStream.F90 is there too now, but it should not offload, because OpenMP does not have `omp target workshare` yet, as far as I know.  It should, but someone has to argue with the `workshare` haters on the committee.

jeffhammond:
I removed the IEEE NaN check from this branch, because I did it wrong.  It is now fixed, and there are other improvements in other branches, which I'll merge after the paper is accepted.  I don't want to break the branch linked in the paper for obvious reasons.

I've now implemented CSV printing, while REAL32 and INT32 options both seem to be working now too.

jeffhammond:
This is ready for review.  I merged in all the good stuff done post-submission.

tomdeakin (approved):

PR #138, opened 2022-08-15
MPI benchmark driver

thomasgibson:
This PR modifies the current driver `main.cpp` and adds MPI support for launching the benchmark across multiple devices. The main takeaways here:
- Each MPI rank is assigned a specific GPU and launches the benchmark
- There is no direct GPU-to-GPU communication happening
  - For the dot-kernel, the resulting sums are reduced across all MPI ranks (on the host) and broadcasted to each rank (via `MPI_Allreduce`).
  - Benchmark error checking is performed on all ranks.
- Measured bandwidths are aggregated across all ranks

The only major question I have is how MPI should be treated by CMake. I am open to suggestions and happy to comply with whatever you all prefer.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -51,6 +51,12 @@
 #include "OMPStream.h"
 #endif
 
+#if USE_MPI
+#include <mpi.h>
+// MPI parameters
+int rank, procs;
+#endif
+
 // Default size of 2^25
 int ARRAY_SIZE = 33554432;
 unsigned int num_times = 100;
@@ -79,21 +85,51 @@ void parseArguments(int argc, char *argv[]);
 
 int main(int argc, char *argv[])
 {
+#if USE_MPI
+  int provided;
+  MPI_Init_thread(&argc, &argv, MPI_THREAD_FUNNELED, &provided);
+  if (provided < MPI_THREAD_FUNNELED) {
+    MPI_Abort(MPI_COMM_WORLD, provided);
+  }
+
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+  MPI_Comm_size(MPI_COMM_WORLD, &procs);
+
+  // Each local rank on a given node will own a single device/GCD
+  MPI_Comm shmcomm;
+  MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0,
+                      MPI_INFO_NULL, &shmcomm);
+  int localRank;
+  MPI_Comm_rank(shmcomm, &localRank);
+  // Set device index to be the local MPI rank
+  deviceIndex = localRank;
+#endif
 
   parseArguments(argc, argv);
 
   if (!output_as_csv)
   {
-    std::cout
-      << "BabelStream" << std::endl
-      << "Version: " << VERSION_STRING << std::endl
-      << "Implementation: " << IMPLEMENTATION_STRING << std::endl;
+#if USE_MPI
+    if (rank == 0) {
+#endif
+      std::cout
+        << "BabelStream" << std::endl
+        << "Version: " << VERSION_STRING << std::endl
+        << "Implementation: " << IMPLEMENTATION_STRING << std::endl;
+#if USE_MPI
+      std::cout << "Number of MPI ranks: " << procs << std::endl;
+    }
+#endif
   }
 
   if (use_float)
     run<float>();
   else
     run<double>(); 
+
+#if USE_MPI
+  MPI_Finalize();
+#endif
 
 }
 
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -79,22 +85,55 @@ void parseArguments(int argc, char *argv[]);
...
+
+  if (provided < MPI_THREAD_FUNNELED)
+    MPI_Abort(MPI_COMM_WORLD, provided);
+
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+  MPI_Comm_size(MPI_COMM_WORLD, &procs);
+
+  // Each rank will run the benchmark on a single device
+  MPI_Comm shared_comm;
+  MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0,
```
tomdeakin, line 102: Why do we need to make a new communicator?

src/main.cpp
```diff
@@ -137,8 +182,15 @@ std::vector<std::vector<double>> run_all(Stream<T> *stream, T& sum)
     timings[3].push_back(std::chrono::duration_cast<std::chrono::duration<double> >(t2 - t1).count());
 
     // Execute Dot
+#if USE_MPI
+    // Synchronize ranks before computing dot-product
+    MPI_Barrier(MPI_COMM_WORLD);
```
tomdeakin, line 187: This overhead is not timed, and waits for 4 kernels to finish before starting the timer for dot. Can you explain the motivation for not including synchronisation time for each kernel?

src/main.cpp
```diff
@@ -201,43 +253,67 @@ void run()
...
-      std::cout << std::setprecision(1) << std::fixed
-                << "Array size: " << ARRAY_SIZE*sizeof(T)*1.0E-6 << " MB"
-                << " (=" << ARRAY_SIZE*sizeof(T)*1.0E-9 << " GB)" << std::endl;
-      std::cout << "Total size: " << 3.0*ARRAY_SIZE*sizeof(T)*1.0E-6 << " MB"
-                << " (=" << 3.0*ARRAY_SIZE*sizeof(T)*1.0E-9 << " GB)" << std::endl;
+      if (mibibytes)
+      {
+        // MiB = 2^20
+        std::cout << std::setprecision(1) << std::fixed
+#if USE_MPI
```
tomdeakin, line 279: There must be a better way to do this, as this is hard to read. Maybe creating a string variable with the label that is initialised depending the MPI case. E.g.,
```
#ifdef USE_MPI
char * array_size_str = "Array size (per rank): ";
#else
char * array_size_str = "Array size: ";
#endif
```

src/main.cpp
```diff
@@ -385,68 +465,90 @@ void run()
       // Calculate average; ignore the first result
       double average = std::accumulate(timings[i].begin()+1, timings[i].end(), 0.0) / (double)(num_times - 1);
 
+      double min = *minmax.first;
+      double max = *minmax.second;
+
+#if USE_MPI
+      MPI_Datatype MPI_DTYPE = use_float ? MPI_FLOAT : MPI_DOUBLE;
+
+      // Collect global min/max timings
```
tomdeakin, line 474: It's not clear that the output will be for a single device, even though this is an MPI code. Is that what we expect, or do we want to (additionally?) report aggregate bandwidth?

src/main.cpp
```diff
@@ -385,68 +465,90 @@ void run()
...
         std::cout
-          << std::left << std::setw(12) << labels[i]
-          << std::left << std::setw(12) << std::setprecision(3) << 
-            ((mibibytes) ? pow(2.0, -20.0) : 1.0E-6) * sizes[i] / (*minmax.first)
-          << std::left << std::setw(12) << std::setprecision(5) << *minmax.first
-          << std::left << std::setw(12) << std::setprecision(5) << *minmax.second
-          << std::left << std::setw(12) << std::setprecision(5) << average
-          << std::endl;
+          << "--------------------------------"
+          << std::endl << std::fixed
```
tomdeakin, line 544: Why do you need to change this? Especially adding the line?

tomdeakin (changes requested):
Please see above for comments on this. I'm also nervous of merging this without some evidence that this works with all supported models.

Can you also think about how we can update the CI to test this is valid?

src/main.cpp
jeffhammond: because of `deviceIndex = localRank;` below.  this is the MPI way to figure out the MPI processes that share a node.  the ranks in this communicator can then be mapped to devices properly, assuming the size of `shared_comm` exceeds the number of devices per node.

tomdeakin:
We've got a large general refactor coming for the main driver coming in #186.
We should also think some more about what bandwidth we expect an MPI+X version should be measuring given there is no communication apart from the dot product. I think we discussed it, but it would be good to document the reasons for wanting MPI+X versions of BabelStream vs running this benchmark on multiple nodes concurrently with `pdsh`, `srun`, etc and post-processing.

PR #139, opened 2022-08-15
Updated hip kernels

thomasgibson:
PR Summary:
- The number of blocks launched for the dot kernel has been modified.
- The partial sums for the dot-kernel is allocated using `hipHostMalloc`, which allocates in a device-visible page. Memory transfer occurs asynchronously and, as a result, requires a `hipDeviceSynchronize` after the kernel is called.
- A number of other minor kernel updates using recommended syntax.

Happy to discuss how much of this you want upstream. For reference (before/after) numbers, here are some quick results in double precision:

**Reference (baseline - develop)**: Ran on an MI-210 with the following arguments with an array size of 2^28 elements: `-s $((2**28))`
```
Copy        1406620.269 0.00305     0.00322     0.00308     
Mul         1404059.445 0.00306     0.00316     0.00312     
Add         1279936.374 0.00503     0.00522     0.00515     
Triad       1272501.462 0.00506     0.00524     0.00516     
Dot         745985.464  0.00576     0.00687     0.00626
```

**Updated dot kernel**:
```
Copy        1406789.357 0.00305     0.00310     0.00307     
Mul         1407020.248 0.00305     0.00622     0.00315     
Add         1279350.509 0.00504     0.00522     0.00515     
Triad       1286205.887 0.00501     0.00525     0.00517     
Dot         1297750.809 0.00331     0.00344     0.00338
```

### DIFF
--- a/src/hip/HIPStream.cpp
+++ b/src/hip/HIPStream.cpp
@@ -23,17 +23,23 @@ void check_error(void)
 
 template <class T>
 HIPStream<T>::HIPStream(const int ARRAY_SIZE, const int device_index)
-  : array_size{ARRAY_SIZE},
-    block_count(array_size / (TBSIZE * elements_per_lane))
 {
 
-  // The array size must be divisible by total number of elements
-  // moved per block for kernel launches
-  if (ARRAY_SIZE % (TBSIZE * elements_per_lane) != 0)
+  // The array size must be divisible by TBSIZE for kernel launches
+  if (ARRAY_SIZE % TBSIZE != 0)
   {
     std::stringstream ss;
-    ss << "Array size must be a multiple of elements operated on per block ("
-       << TBSIZE * elements_per_lane
+    ss << "Array size must be a multiple of " << TBSIZE;
+    throw std::runtime_error(ss.str());
+  }
+
+  // The array size must be divisible by total number of elements
+  // moved per block for the dot kernel
+  if (ARRAY_SIZE % (TBSIZE * dot_elements_per_lane) != 0)
+  {
+    std::stringstream ss;
+    ss << "Array size for the dot kernel must be a multiple of elements operated on per block ("
+       << TBSIZE * dot_elements_per_lane
        << ").";
     throw std::runtime_error(ss.str());
   }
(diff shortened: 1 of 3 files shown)


tomdeakin:
Thanks for the PR. In general it looks OK, but I would like more details on the benefit for manually strip-mining the loop in the `nstream` kernel. I don't think we do this for any other model - why do we need to do it for HIP / AMD GPUs?

thomasgibson:
Thanks for your comment!

> Thanks for the PR. In general it looks OK, but I would like more details on the benefit for manually strip-mining the loop in the nstream kernel. I don't think we do this for any other model - why do we need to do it for HIP / AMD GPUs?

It's not strictly necessary, this was mostly a result of experimentation to generated vector load instructions from device memory in an attempt to boost the bandwidth performance of the kernel. This had a consistent, albeit minor, increase in bandwidth performance for n-stream. This was mainly useful during experimentation, but was hesitant to modify it any further as I recall from our last conversation that we want to avoid significantly "ninja'd" code in the main repo.

This was more impactful in older versions of ROCm, but since 5.2 we're not seeing a huge benefit from doing this. This is probably for the best, since our compilers are generating better instructions from simpler code. There's still some room for improvement and this kernel in particular is one I'm looking at including in our HBM stressors/benchmarks. I'm happy to leave it mostly the way it is. Will update this PR accordingly.

src/hip/HIPStream.cpp
```diff
@@ -33,6 +33,17 @@ HIPStream<T>::HIPStream(const int ARRAY_SIZE, const int device_index)
     throw std::runtime_error(ss.str());
   }
 
+  // The array size must be divisible by total number of elements
```
tomdeakin, line 36: Why do we need to add this restriction? The CUDA version handles this case.

src/hip/HIPStream.cpp
```diff
@@ -186,32 +198,32 @@ template <typename T>
...
   __shared__ T tb_sum[TBSIZE];
 
-  int i = hipBlockDim_x * hipBlockIdx_x + hipThreadIdx_x;
-  const size_t local_i = hipThreadIdx_x;
+  const size_t local_i = threadIdx.x;
+  size_t i = blockDim.x * blockIdx.x + local_i;
 
   tb_sum[local_i] = 0.0;
-  for (; i < array_size; i += hipBlockDim_x*hipGridDim_x)
+  for (; i < array_size; i += blockDim.x*gridDim.x)
```
tomdeakin, line 223: Indeed - this loop here should mean that the kernel is valid even if the block size does not evenly divide...

tomdeakin (changes requested):

src/hip/HIPStream.cpp
arghdos: I think you're right Tom, this is a hold-over from a previous version I think.

thomasgibson: @tomdeakin Apologies for the late reply --- I've gone ahead and pushed an additional commit removing this

tomdeakin (approved):
Thanks for contributing this, happy to merge this now. Thanks for the effort!

Issue #142, opened 2022-09-23
support NVC++ CUDA compiler


jeffhammond:
NVC++ supports CUDA, although it is not 100% compatible with NVCC due to C++ legalese.

The following is the required source code change for Thrust.  I assume the one for CUDA is similar.

There also need to be CMake changes, to add the `-cuda` flag and recognize `-DCMAKE_CUDA_COMPILER=nvc++`.

```patch
index 3a57ab0..f15a392 100644
--- a/src/thrust/ThrustStream.cu
+++ b/src/thrust/ThrustStream.cu
@@ -145,7 +145,7 @@ T ThrustStream<T>::dot()
 #if THRUST_DEVICE_SYSTEM == THRUST_DEVICE_SYSTEM_CUDA || \
     (defined(THRUST_DEVICE_SYSTEM_HIP) && THRUST_DEVICE_SYSTEM_HIP == THRUST_DEVICE_SYSTEM)

-#ifdef __NVCC__
+#if defined(__NVCC__) || defined(__NVCOMPILER_CUDA__)
 #define IMPL_FN__(fn) cuda ## fn
 #define IMPL_TYPE__(tpe) cuda ## tpe
 #elif defined(__HIP_PLATFORM_HCC__)
```

PR #144, opened 2022-09-26
accept NVHPC NVC++ as a CUDA compiler when it is so

jeffhammond:
fixes https://github.com/UoB-HPC/BabelStream/issues/142 except for the CMake part, which i attempted in vain to address.

Signed-off-by: Jeff Hammond <jehammond@nvidia.com>


tom91136:
LGTM, thanks!

PR #159, opened 2023-08-01
Support for other microarchitectures

pranav-sivaraman:
The default release flags `-march=native` can conflict when building for other architectures such as PowerPC. Mainly would allow the spack package to inject architecture specific [flags](https://spack.readthedocs.io/en/latest/basic_usage.html#support-for-specific-microarchitectures) without any conflicts.

This change also can just be a patch on the spack package instead of a merge to the original repo. Not sure what the best approach would be.


gonzalobg:
@pranav-sivaraman If we were to provide a `-DMARCH_NATIVE=Off` option, that spack could set, would that work for you?

tomdeakin:
I want the build system we include here to work on its own without a critical dependency on a package manager, and so we need to have `-march=native` included as a default so that the code builds sensibly out of the box.

At the same time, I also want to support Spack, and we're upstreaming a big update to the BabelStream Spack package (there is a PR open). We still want to have some resolution to this so Spack can support cross-compiling.

Another way to resolve this is through the Spack package directly. The package could tweak the flags set by our CMake build system. Spack supports that, and does it for plenty of packages. So, I wonder if that is a better approach.

RFC!

gonzalobg:
My proposal is to add a new option, and protect flags for native builds with them, enabling them by default, but also allowing spack and others to override them by disabling the option:

```
option(BUILD_NATIVE "Builds for the current systems CPU and GPU architecture." ON)

if (BUILD_NATIVE)
    set(DEFAULT_RELEASE_FLAGS -O3 -march=native)
    # Set -mcpu=native on Arm
    # Set -arch=native, -gpu=ccnative, ... on GPUs
endif()
```

I'm not a spack expert, so don't know if this would solve the issue on their end or not.

pranav-sivaraman:
> My proposal is to add a new option, and protect flags for native builds with them, enabling them by default, but also allowing spack and others to override them by disabling the option:

That should be fine. The Spack package can disable this option by default as it will already inject microarchitecture flags automatically, so there should be no conflicts. The option can also be extracted as a patch and backported to older versions into the Spack package. However, there are places in the CMake where CMAKE_CXX_COMPILER is set (I think just in SYCL), which overrides the Spack compiler wrappers that inject rpath and microarchitecture flags. Those need some special handling if the defaults are to be kept.

PR #160, opened 2023-08-24
Delete make.inc.* Files for Spack

kaanolgu:
After the https://github.com/spack/spack/pull/38559 merges, we will not need the make.inc.* files since they will be generated by Spack


tomdeakin:
I think we'll keep those files for now, as there is still a use in having BabelStream build outside of Spack.

PR #172, opened 2023-11-05
Update README.md to Address Changes in SYCL2020 Renaming

kaanolgu:
The sycl2020 model variant name needs an update to comply with naming scheme in Babelstream 5.0


tom91136:
Thanks, CI failed for unrelated space reasons (this has happened before with PRs from other branches so there might be a quota somewhere)

Issue #113, opened 2021-11-30
Publish JuliaStream.jl


tom91136:
So that the implementation can be used from the REPL directly.

tom91136:
TODO: use template `@ JuliaRegistrator register subdir=src/julia/JuliaStream.jl branch=main`  after everything is merged to main

tom91136:
@JuliaRegistrator register subdir=src/julia/JuliaStream.jl branch=main

tom91136:
@tomdeakin Can you set me as public please, thanks.

tom91136:
@JuliaRegistrator register subdir=src/julia/JuliaStream.jl branch=main

tom91136:
OK, so based on the guidelines, we need to:
 - [ ] Renamed JuliaStream.jl to BabelStream.jl
 - [ ] Copy licence from top level and set it in the project as well
 - [ ] Fix up `[compat]` entries
 - [ ] Set the correct version

PR #175, opened 2023-11-14
Rename Julia package to `BabelStream.jl` and add compats

giordano:
Ref: https://github.com/UoB-HPC/BabelStream/issues/113#issuecomment-1760395399.

Note: I haven't copied the license because there's a problem: the Julia General registry [requires, without exceptions, an OSI-approved license](https://github.com/JuliaRegistries/General/blob/69df77f5c026e8428c72803f4130b6f993e18d48/README.md#registering-a-package-in-general) that can be automatically detected by [licensecheck](https://github.com/google/licensecheck), but the [top-level license](https://github.com/UoB-HPC/BabelStream/blob/8389fc43a327e23e72611957cc92d4586e85352c/LICENSE) in this repository is a custom one, with some strong restrictions.  I don't know if there's any appetite for using a standard OSI license for the Julia implementation only, if not, I'm afraid registration in General will not be possible, but this package can always be installed with `Pkg.add(; url="https://github.com/UoB-HPC/BabelStream", subdir="...")`.  Side note, the top-level license references the wikipage http://github.com/UoB-HPC/BabelStream/wiki/Run-Rules which doesn't exist.

CC: @tom91136.

TODO:

* [x] Rename JuliaStream BabelStream everywhere


tom91136:
@tomdeakin Could we sub-license or dual licence only the Julia port (or BabelStream itself)?

tomdeakin:
The license is based on STREAM, and the restrictions there and here are reasonable and important. I don't want to change the license at this time. I'll fix the Wiki link though - thanks for pointing that out.

tom91136:
The suggested `Pkg.add(...)` installation method works well for the intended benchmark use case. Thanks again for the PR, we'll merge this and document the required steps for a benchmark run.

tom91136:
- [ ] TODO to self: Add a few sentences documenting the installation procedure

tomdeakin:
@giordano @tom91136 - see request above.

PR #180, opened 2023-12-04
Remove AdaptiveCpp workaround in dpl_shim.h to allow for automatic pefetch

illuhad:
To my knowledge, the AdaptiveCpp-specific workaround in `dpl_shim.h` is no longer necessary as the linking issue was fixed some time ago.

Additionally, this workaround prevents a new optmization in AdaptiveCpp stdpar from taking place: AdaptiveCpp stdpar can now emit automatic data prefetches. But this only works, if the data is managed by the stdpar runtime, which tracks pointers it knows. If sycl::malloc_shared is used directly, the stdpar optimizations do not recognize the pointer and won't apply the optimization.
(the reasoning being that if someone calls SYCL functionality directly, they might want to exert more control and emit prefetches or mem_advise operations by themselves).

Other projects which also use this header like cloverleaf are also affected. Will create a PR there too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,13 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
+
 ### Removed
 - Remove support for ComputeCpp compiler
+
+### Changed
+- Removed std-data/std-indices memory management workaround for AdaptiveCpp
+
 
 ## [v5.0] - 2023-10-12
 ### Added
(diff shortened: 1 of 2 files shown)


illuhad:
@tomdeakin Is this okay now? It addresses your feedback from #177.

tomdeakin:
Looks good, thanks @illuhad. As long as the CI comes back OK I'll merge. Thanks!

tomdeakin (approved):

illuhad:
CI looks good to me @tomdeakin  :)

PR #184, opened 2024-01-26
set device_id for Kokkos

milthorpe:
Previously, the Kokkos implementation ignores the command line parameter `--device` and always runs using the default device. This change initializes Kokkos with the specified device.


tomdeakin:
Looks good, just re-running the CI.

tomdeakin:
CI is failing only after the cpp test passes, so this looks good to merge. Thanks for the contribution!

PR #185, opened 2024-02-27
Support array_sizes that are not divisible by the thread block size for CUDA and HIP

gonzalobg:
Currently, CUDA and HIP do not support array sizes that are not divisible by the thread block size, Since most other programming models support these, this PR removes this limitation to improve BabelStream's fairness.

This is done using a technique that is widely used in practice, commonly known as a "grid-strided loop" (see [CUDA Pro Tip: Write Flexible Kernels with Grid-Stride Loops](https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/)). The number of thread blocks in the launch configuration is rounded up to cover all array elements using `ceil_div`.

This PR also updates the CUDAStream implementation to modern CUDA C++ practices by simplifying error handling, and using an explicit `cudaStream_t`. The impact of these two changes on performance is negligible; they are only cosmetic changes.

### DIFF
--- a/src/cuda/CUDAStream.cu
+++ b/src/cuda/CUDAStream.cu
@@ -16,6 +16,7 @@
 // The do while is there to make sure you remember to put a semi-colon after calling CU
 #define CU(EXPR) do { auto __e = (EXPR); if (__e != cudaSuccess) error(__FILE__, __LINE__, #EXPR, __e); } while(false)
 
+// It is best practice to include __device__ and constexpr even though in BabelStream it only needs to be __host__ const
 __host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
 
 cudaStream_t stream;
--- a/src/cuda/CUDAStream.cu
+++ b/src/cuda/CUDAStream.cu
@@ -15,6 +15,8 @@
 
 #define CU(EXPR) if (auto __e = (EXPR); __e != cudaSuccess) error(__FILE__, __LINE__, #EXPR, __e);
 
+__host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
+
 template <class T>
 CUDAStream<T>::CUDAStream(const int ARRAY_SIZE, const int device_index)
 {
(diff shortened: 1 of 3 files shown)


tom91136:
LGTM, cc @tomdeakin

src/cuda/CUDAStream.cu
```diff
@@ -8,36 +8,28 @@
...
-    exit(err);
-  }
+[[noreturn]] inline void error(char const* file, int line, char const* expr, cudaError_t e) {
+  std::fprintf(stderr, "Error at %s:%d: %s (%d)\n  %s\n", file, line, cudaGetErrorString(e), e, expr);
+  exit(e);
 }
 
+#define CU(EXPR) do { auto __e = (EXPR); if (__e != cudaSuccess) error(__FILE__, __LINE__, #EXPR, __e); } while(false)
+
+__host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
```
tomdeakin, line 20:
```suggestion
// It is best practice to include __device__ and constexpr even though in BabelStream it only needs to be __host__ const
__host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
```

src/cuda/CUDAStream.cu
```diff
@@ -8,36 +8,28 @@
...
-  {
-    std::cerr << "Error: " << cudaGetErrorString(err) << std::endl;
-    exit(err);
-  }
+[[noreturn]] inline void error(char const* file, int line, char const* expr, cudaError_t e) {
+  std::fprintf(stderr, "Error at %s:%d: %s (%d)\n  %s\n", file, line, cudaGetErrorString(e), e, expr);
+  exit(e);
 }
 
+#define CU(EXPR) do { auto __e = (EXPR); if (__e != cudaSuccess) error(__FILE__, __LINE__, #EXPR, __e); } while(false)
```
tomdeakin, line 17:
```suggestion
// The do while is there to make sure you remember to put a semi-colon after calling CU
#define CU(EXPR) do { auto __e = (EXPR); if (__e != cudaSuccess) error(__FILE__, __LINE__, #EXPR, __e); } while(false)
```

src/hip/HIPStream.cpp
```diff
@@ -21,18 +21,11 @@ void check_error(void)
   }
 }
 
+__host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
```
tomdeakin, line 25:
```suggestion
// It is best practice to include __device__ and constexpr even though in BabelStream it only needs to be __host__ const
__host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
```

PR #186, opened 2024-03-08
Improvements to main.cpp

gonzalobg:
This PR gives a pass to `main.cpp` to make it much easier to add newer benchmarks in the future.

Main changes are:
- `--only` option enables specifying different groups or individual benchmarks, e.g., `--only Classic` runs the classic 5 benchmarks, `--only All` runs all benchmarks (e.g. including nstream), and `--only Triad` runs only triad, etc.
- `--triad-only`/`--nstream-only` options removed (replaced by `--only Triad` and `--only Nstream`).
- `--only Triad` now runs `triad` in the exact same way as all other benchmarks, i.e., timing each individual iteration. Before, the outer loop was being timed.
- `--gigabytes` and similar option print the bandwidths in GB/s (or GiB/s, MB/s, MiB/s)

Bug fixes:
- Metrics for `Init` were incorrect; the timers from `Read` were incorrectly used for both `Init` and `Read`.

Future:
- After these changes, adding new benchmarks like `Scan` is pretty straightforward: bump benchmark count by 1, and add them to all places in the code that fail (runner, checker, and that's it..).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,8 +31,8 @@ endmacro()
 # the final executable name
 set(EXE_NAME babelstream)
 
-# for chrono and some basic CXX features, models can overwrite this if required
-set(CMAKE_CXX_STANDARD 11)
+# for chrono, some basic CXX features, and generic lambdas; models can overwrite this if required
+set(CMAKE_CXX_STANDARD 14)
 
 if (NOT CMAKE_BUILD_TYPE)
     message("No CMAKE_BUILD_TYPE specified, defaulting to 'Release'")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,8 +31,8 @@ endmacro()
 # the final executable name
 set(EXE_NAME babelstream)
 
-# for chrono and some basic CXX features, models can overwrite this if required
-set(CMAKE_CXX_STANDARD 11)
+# for chrono, make_unique, and some basic CXX features, models can overwrite this if required
+set(CMAKE_CXX_STANDARD 14)
 
 if (NOT CMAKE_BUILD_TYPE)
     message("No CMAKE_BUILD_TYPE specified, defaulting to 'Release'")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,8 +31,8 @@ endmacro()
 # the final executable name
 set(EXE_NAME babelstream)
 
-# for chrono, some basic CXX features, and generic lambdas; models can overwrite this if required
-set(CMAKE_CXX_STANDARD 14)
+# for chrono and some basic CXX features, models can overwrite this if required
+set(CMAKE_CXX_STANDARD 11)
 
 if (NOT CMAKE_BUILD_TYPE)
     message("No CMAKE_BUILD_TYPE specified, defaulting to 'Release'")
--- a/src/StreamModels.h
+++ b/src/StreamModels.h
@@ -36,66 +36,66 @@
 #endif
 
 template <typename T>
-std::unique_ptr<Stream<T>> make_stream(int ARRAY_SIZE, unsigned int deviceIndex) {
+std::unique_ptr<Stream<T>> make_stream(int array_size, int deviceIndex) {
 #if defined(CUDA)
   // Use the CUDA implementation
-  return std::make_unique<CUDAStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<CUDAStream<T>>(array_size, deviceIndex);
 
 #elif defined(HIP)
   // Use the HIP implementation
-  return std::make_unique<HIPStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<HIPStream<T>>(array_size, deviceIndex);
 
 #elif defined(HC)
   // Use the HC implementation
-  return std::make_unique<HCStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<HCStream<T>>(array_size, deviceIndex);
 
 #elif defined(OCL)
   // Use the OpenCL implementation
-  return std::make_unique<OCLStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<OCLStream<T>>(array_size, deviceIndex);
 
 #elif defined(USE_RAJA)
   // Use the RAJA implementation
-  return std::make_unique<RAJAStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<RAJAStream<T>>(array_size, deviceIndex);
 
 #elif defined(KOKKOS)
   // Use the Kokkos implementation
-  return std::make_unique<KokkosStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<KokkosStream<T>>(array_size, deviceIndex);
 
 #elif defined(STD_DATA)
   // Use the C++ STD data-oriented implementation
-  return std::make_unique<STDDataStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<STDDataStream<T>>(array_size, deviceIndex);
 
 #elif defined(STD_INDICES)
   // Use the C++ STD index-oriented implementation
-  return std::make_unique<STDIndicesStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<STDIndicesStream<T>>(array_size, deviceIndex);
 
 #elif defined(STD_RANGES)
   // Use the C++ STD ranges implementation
-  return std::make_unique<STDRangesStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<STDRangesStream<T>>(array_size, deviceIndex);
 
 #elif defined(TBB)
   // Use the C++20 implementation
-  return std::make_unique<TBBStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<TBBStream<T>>(array_size, deviceIndex);
 
 #elif defined(THRUST)
   // Use the Thrust implementation
-  return std::make_unique<ThrustStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<ThrustStream<T>>(array_size, deviceIndex);
 
 #elif defined(ACC)
   // Use the OpenACC implementation
-  return std::make_unique<ACCStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<ACCStream<T>>(array_size, deviceIndex);
 
 #elif defined(SYCL) || defined(SYCL2020)
   // Use the SYCL implementation
-  return std::make_unique<SYCLStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<SYCLStream<T>>(array_size, deviceIndex);
 
 #elif defined(OMP)
   // Use the OpenMP implementation
-  return std::make_unique<OMPStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<OMPStream<T>>(array_size, deviceIndex);
 
 #elif defined(FUTHARK)
   // Use the Futhark implementation
-  return std::make_unique<FutharkStream<T>>(ARRAY_SIZE, deviceIndex);
+  return std::make_unique<FutharkStream<T>>(array_size, deviceIndex);
 
 #else
 
--- a/src/StreamModels.h
+++ b/src/StreamModels.h
@@ -36,7 +36,7 @@
 #endif
 
 template <typename T>
-std::unique_ptr<Stream<T>> construct_stream(int ARRAY_SIZE, unsigned int deviceIndex) {
+std::unique_ptr<Stream<T>> make_stream(int ARRAY_SIZE, unsigned int deviceIndex) {
 #if defined(CUDA)
   // Use the CUDA implementation
   return std::make_unique<CUDAStream<T>>(ARRAY_SIZE, deviceIndex);
--- a/src/StreamModels.h
+++ b/src/StreamModels.h
@@ -1,0 +1,105 @@
+#pragma once
+#include <memory>
+
+#if defined(CUDA)
+#include "CUDAStream.h"
+#elif defined(STD_DATA)
+#include "STDDataStream.h"
+#elif defined(STD_INDICES)
+#include "STDIndicesStream.h"
+#elif defined(STD_RANGES)
+#include "STDRangesStream.hpp"
+#elif defined(TBB)
+#include "TBBStream.hpp"
+#elif defined(THRUST)
+#include "ThrustStream.h"
+#elif defined(HIP)
+#include "HIPStream.h"
+#elif defined(HC)
+#include "HCStream.h"
+#elif defined(OCL)
+#include "OCLStream.h"
+#elif defined(USE_RAJA)
+#include "RAJAStream.hpp"
+#elif defined(KOKKOS)
+#include "KokkosStream.hpp"
+#elif defined(ACC)
+#include "ACCStream.h"
+#elif defined(SYCL)
+#include "SYCLStream.h"
+#elif defined(SYCL2020)
+#include "SYCLStream2020.h"
+#elif defined(OMP)
+#include "OMPStream.h"
+#elif defined(FUTHARK)
+#include "FutharkStream.h"
+#endif
+
+template <typename T>
+std::unique_ptr<Stream<T>> construct_stream(int ARRAY_SIZE, unsigned int deviceIndex) {
+#if defined(CUDA)
+  // Use the CUDA implementation
+  return std::make_unique<CUDAStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(HIP)
+  // Use the HIP implementation
+  return std::make_unique<HIPStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(HC)
+  // Use the HC implementation
+  return std::make_unique<HCStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(OCL)
+  // Use the OpenCL implementation
+  return std::make_unique<OCLStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(USE_RAJA)
+  // Use the RAJA implementation
+  return std::make_unique<RAJAStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(KOKKOS)
+  // Use the Kokkos implementation
+  return std::make_unique<KokkosStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(STD_DATA)
+  // Use the C++ STD data-oriented implementation
+  return std::make_unique<STDDataStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(STD_INDICES)
+  // Use the C++ STD index-oriented implementation
+  return std::make_unique<STDIndicesStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(STD_RANGES)
+  // Use the C++ STD ranges implementation
+  return std::make_unique<STDRangesStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(TBB)
+  // Use the C++20 implementation
+  return std::make_unique<TBBStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(THRUST)
+  // Use the Thrust implementation
+  return std::make_unique<ThrustStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(ACC)
+  // Use the OpenACC implementation
+  return std::make_unique<ACCStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(SYCL) || defined(SYCL2020)
+  // Use the SYCL implementation
+  return std::make_unique<SYCLStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(OMP)
+  // Use the OpenMP implementation
+  return std::make_unique<OMPStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(FUTHARK)
+  // Use the Futhark implementation
+  return std::make_unique<FutharkStream<T>>(ARRAY_SIZE, deviceIndex);
+
+#else
+
+  #error unknown benchmark
+
+#endif
+}
--- a/src/Unit.h
+++ b/src/Unit.h
@@ -1,51 +1,31 @@
 #pragma once
 #include <iostream>
 
-enum { MegaByte, GigaByte, MibiByte, GibiByte };
-
 // Units for output:
 struct Unit {
-  int value;
-  Unit(int v) : value(v) {}
-  double fmt(double bytes) {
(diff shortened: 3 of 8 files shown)


src/main.cpp
```diff
@@ -1,82 +1,92 @@
...
-#elif defined(SYCL)
-#include "SYCLStream.h"
-#elif defined(SYCL2020)
-#include "SYCLStream2020.h"
-#elif defined(OMP)
-#include "OMPStream.h"
-#elif defined(FUTHARK)
-#include "FutharkStream.h"
-#endif
+using namespace std;
```
tom91136: Let's not do this :)

CMakeLists.txt
```diff
@@ -31,8 +31,8 @@ endmacro()
 # the final executable name
 set(EXE_NAME babelstream)
 
-# for chrono and some basic CXX features, models can overwrite this if required
-set(CMAKE_CXX_STANDARD 11)
+# for chrono, some basic CXX features, and generic lambdas; models can overwrite this if required
+set(CMAKE_CXX_STANDARD 14)
```
tom91136, line 35: I don't think we ended up using generic lambdas in the PR.

src/main.cpp
```diff
@@ -1,82 +1,92 @@
...
+  if (id == 3 && selection == Benchmark::Triad)   return true;
+  if (id == 4 && selection == Benchmark::Dot)     return true;
+  if (id == 5 && selection == Benchmark::Nstream) return true;
+  return false;
+}
+
+// Prints all available benchmark labels:
+template <typename OStream>
+void print_labels(OStream& os) {
+  for (int i = 0; i < num_benchmarks; ++i) {
```
tom91136: `size_t i`

src/main.cpp
```diff
@@ -520,57 +279,54 @@ void check_solution(const unsigned int ntimes, std::vector<T>& a, std::vector<T>
 
   const T scalar = startScalar;
 
-  for (unsigned int i = 0; i < ntimes; i++)
+  for (int b = 0; b < num_benchmarks; ++b)
```
tom91136: `size_t b`

src/main.cpp
```diff
@@ -1,82 +1,92 @@
...
+void print_labels(OStream& os) {
+  for (int i = 0; i < num_benchmarks; ++i) {
+    os << labels[i];
+    if (i != (num_benchmarks - 1)) os << ",";
+  }
+}
+
+// Clock and duration types:
+using clk_t = chrono::high_resolution_clock;
+using dur_t = chrono::duration<double>;
```
tom91136: These two are not used that frequently in the new structure so for clarity just use without alias in `time`?

src/main.cpp
```diff
@@ -520,57 +279,54 @@ void check_solution(const unsigned int ntimes, std::vector<T>& a, std::vector<T>
...
-    } else if (selection == Benchmark::Triad)
-    {
-      goldA = goldB + scalar * goldC;
-    } else if (selection == Benchmark::Nstream)
+    if (!run_benchmark(b)) continue;
+
+    for (unsigned int i = 0; i < ntimes; i++)
     {
-      goldA += goldB + scalar * goldC;
+      switch((Benchmark)b) {
```
tom91136: `static_cast<Benchmark>(b)`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+#include <iostream>
+
+enum { MegaByte, GigaByte, MibiByte, GibiByte };
+
+// Units for output:
+struct Unit {
+  int value;
+  Unit(int v) : value(v) {}
```
tom91136: `explicit Unit(int v)...`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
...
+  double fmt(double bytes) {
+    switch(value) {
+    case MibiByte: return pow(2.0, -20.0) * bytes;
+    case MegaByte: return 1.0E-6 * bytes;
+    case GibiByte: return pow(2.0, -30.0) * bytes;
+    case GigaByte: return 1.0E-9 * bytes;
```
tom91136: Qualify math operations

src/Unit.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+#include <iostream>
+
+enum { MegaByte, GigaByte, MibiByte, GibiByte };
```
tom91136: While we're here, we may want to add Terabyte/Petabyte as we're getting really close to having values joining together in the CSV format.

src/StreamModels.h
```diff
@@ -0,0 +1,105 @@
...
+#elif defined(SYCL2020)
+#include "SYCLStream2020.h"
+#elif defined(OMP)
+#include "OMPStream.h"
+#elif defined(FUTHARK)
+#include "FutharkStream.h"
+#endif
+
+template <typename T>
+std::unique_ptr<Stream<T>> construct_stream(int ARRAY_SIZE, unsigned int deviceIndex) {
```
tom91136: Bike-shed: `make_stream`?

src/Unit.h
tom91136: Use scoped enums

src/Unit.h
```diff
@@ -0,0 +1,51 @@
...
+  double fmt(double bytes) {
+    switch(value) {
+    case MibiByte: return pow(2.0, -20.0) * bytes;
+    case MegaByte: return 1.0E-6 * bytes;
+    case GibiByte: return pow(2.0, -30.0) * bytes;
+    case GigaByte: return 1.0E-9 * bytes;
+    default: std::cerr << "Unimplemented!" << std::endl; abort();
+    }
+  }
+  char const* str() {
```
tom91136: `str() const {`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
...
+    switch(value) {
+    case MibiByte: return "MiB";
+    case MegaByte: return "MB";
+    case GibiByte: return "GiB";
+    case GigaByte: return "GB";
+    default: std::cerr << "Unimplemented!" << std::endl; abort();
+    }
+  }
+  Unit kibi() {
```
tom91136: `const`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
...
+    switch(value) {
+    case MegaByte: return Unit(MibiByte);
+    case GigaByte: return Unit(GibiByte);
+    default: return *this;
+    }
+  }
+  Unit byte() {
```
tom91136: `const`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
...
+    switch(value) {
+    case MibiByte: return Unit(MegaByte);
+    case GibiByte: return Unit(GigaByte);
+    default: return *this;
+    }
+  }
+  char const* lower() {
```
tom91136: `const`

src/Unit.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+#include <iostream>
+
+enum { MegaByte, GigaByte, MibiByte, GibiByte };
+
+// Units for output:
+struct Unit {
+  int value;
+  Unit(int v) : value(v) {}
+  double fmt(double bytes) {
```
tom91136: `const`

src/main.cpp
```diff
@@ -1,82 +1,92 @@
...
+string csv_separator = ",";
 
-template <typename T>
-void run();
+// Benchmarks:
+constexpr size_t num_benchmarks = 6;
+array<char const*, num_benchmarks> labels = {"Copy", "Add", "Mul", "Triad", "Dot", "Nstream"};
+// Weights data moved by benchmark & therefore achieved BW:
+// bytes = weight * sizeof(T) * ARRAY_SIZE -> bw = bytes / dur
+array<size_t, num_benchmarks> weight = {/*Copy:*/ 2, /*Add:*/ 2, /*Mul:*/ 3, /*Triad:*/ 3, /*Dot:*/ 2, /*Nstream:*/ 4};
```
tom91136: It might be worth having a `struct Benchmark { std::string label; size_t weight;  }` and a `std::unordered_map` for the lookup or something.

src/main.cpp
```diff
@@ -86,431 +96,180 @@ int main(int argc, char *argv[])
...
+      << "Implementation: " << IMPLEMENTATION_STRING << endl;
   }
 
   if (use_float)
     run<float>();
   else
-    run<double>(); 
+    run<double>();
 
+  return 0;
```
tom91136: `return EXIT_SUCCESS;`

tom91136 (changes requested):
Minor nitpicks and one major change (no `using namespace std;`)

gonzalobg:
EDIT: will do this in a sub-sequent PR using `intptr_t`

I've noticed while testing on GPUs with more than 104 GB of memory that the results were "wrong".
The problem is that with `int array_size`, the largest array is 2^31 * 8 * 1e-9 = ~17.2 GB (51.6 GB with 3 arrays). With `unsigned int` it was ~104 GB. When trying to use larger arrays, validation would fail due to overflow.

Fixing this required using `size_t` in a couple of places. I've verified this for `cuda`, `thrust`, `std-par`, openmp, etc. I've turned the check solutions "on by default", adding a `--silence-errors` flag that one can pass to continue the benchmark even if validation fails.

src/main.cpp
gonzalobg: Have added the `struct`, but for now the `array` still seems fine.

src/cuda/CUDAStream.cu
```diff
@@ -231,50 +222,33 @@ void CUDAStream<T>::nstream()
...
-    if (local_i < offset)
-    {
-      tb_sum[local_i] += tb_sum[local_i+offset];
-    }
+    if (tidx < offset) smem[tidx] += smem[tidx+offset];
   }
 
-  if (local_i == 0)
-    sum[blockIdx.x] = tb_sum[local_i];
+  if (tidx == 0) sums[blockIdx.x] = smem[tidx];
```
tomdeakin, line 241:
```suggestion
  // First thread writes to host memory directly from the device
  if (tidx == 0) sums[blockIdx.x] = smem[tidx];
```

src/Unit.h
```diff
@@ -0,0 +1,31 @@
...
+    case Kind::MibiByte: return std::pow(2.0, -20.0) * bytes;
+    case Kind::MegaByte: return 1.0E-6 * bytes;
+    case Kind::GibiByte: return std::pow(2.0, -30.0) * bytes;
+    case Kind::GigaByte: return 1.0E-9 * bytes;
+    case Kind::TebiByte: return std::pow(2.0, -40.0) * bytes;
+    case Kind::TeraByte: return 1.0E-12 * bytes;      
+    default: std::cerr << "Unimplemented!" << std::endl; std::abort();
+    }
+  }
+  char const* str() const {
```
tomdeakin, line 24:
```suggestion

  char const* str() const {
```

src/Unit.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <iostream>
+
+// Units for output:
+struct Unit {
+  enum class Kind { MegaByte, GigaByte, TeraByte, MibiByte, GibiByte, TebiByte };
+  Kind value;
+  explicit Unit(Kind v) : value(v) {}
+  double fmt(double bytes) const {
```
tomdeakin, line 12:
```suggestion

  double fmt(double bytes) const {
```

src/Unit.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+#include <iostream>
+
+// Units for output:
+struct Unit {
+  enum class Kind { MegaByte, GigaByte, TeraByte, MibiByte, GibiByte, TebiByte };
+  Kind value;
```
tomdeakin, line 8:
```suggestion

  Kind value;
  
```

src/main.cpp
```diff
@@ -1,87 +1,80 @@
...
 // Default size of 2^25
 int ARRAY_SIZE = 33554432;
-unsigned int num_times = 100;
-unsigned int deviceIndex = 0;
+size_t num_times = 100;
+size_t deviceIndex = 0;
 bool use_float = false;
 bool output_as_csv = false;
-bool mibibytes = false;
+Unit unit{Unit::Kind::MegaByte};
```
tomdeakin, line 31:
```suggestion
// Default unit of memory is MegaBytes (as per STREAM) 
Unit unit{Unit::Kind::MegaByte};
```

src/main.cpp
```diff
@@ -1,87 +1,81 @@
...
+// Benchmark Identifier: identifies individual & groups of benchmarks:
+// - Classic: 5 classic kernels: Copy, Mul, Add, Triad, Dot.
+// - All: all kernels.
+// - Individual kernels only.
+enum class BenchId : int {Copy, Mul, Add, Triad, Nstream, Dot, Classic, All};
+
+struct Benchmark {
+  BenchId id;
+  char const* label;
+  // Weights data moved by benchmark & therefore achieved BW:
```
tomdeakin:
```suggestion
  // Weight counts data elements of original arrays moved each loop iteration - used to calculate achieved BW:
```

src/main.cpp
```diff
@@ -95,106 +89,59 @@ int main(int argc, char *argv[])
...
+    case BenchId::Nstream: return time([&] { stream->nstream(); });
+    default:
+      std::cerr << "Unimplemented benchmark: " << b.label << std::endl;
+      abort();
+    }
+  };
 
   // Main loop
-  for (unsigned int k = 0; k < num_times; k++)
+  for (size_t i = 0; i < num_benchmarks; ++i)
```
tomdeakin: I think this is the main outstanding change on this PR. We need the classic kernels (`BenchId::Classic`) to run as:
```
for (size_t k = 0; k < num_times; k++) {
  for (size_t i = 0; i < num_benchmarks; ++i) {
    dt(bench[i])
  }
}
```
rather than the current order.
Therefore, the correct ordering runs:
`Copy -> Mul -> Add -> Triad -> Dot -> Copy -> Multiple -> ...`

src/main.cpp
```diff
@@ -203,393 +150,226 @@ void run()
...
-      goldB = scalar * goldC;
-      goldC = goldA + goldB;
-      goldA = goldB + scalar * goldC;
-    } else if (selection == Benchmark::Triad)
-    {
-      goldA = goldB + scalar * goldC;
-    } else if (selection == Benchmark::Nstream)
+    if (!run_benchmark(bench[b])) continue;
+
+    for (size_t i = 0; i < num_times; i++)
```
tomdeakin: As above, I think the loop order needs switching for `check_solution()` does the arithmetic in the same order.

src/main.cpp
```diff
@@ -203,393 +150,226 @@ void run()
...
     {
-      goldA += goldB + scalar * goldC;
+      switch(bench[b].id) {
+      case BenchId::Copy:    goldC = goldA; break;
+      case BenchId::Mul:     goldB = scalar * goldC; break;
+      case BenchId::Add:     goldC = goldA + goldB; break;
+      case BenchId::Triad:   goldA = goldB + scalar * goldC; break;
+      case BenchId::Nstream: goldA += goldB + scalar * goldC; break;
+      case BenchId::Dot:     goldS = goldA * goldB * T(ARRAY_SIZE); break;
```
tomdeakin:
```suggestion
      case BenchId::Dot:     goldS = goldA * goldB * T(ARRAY_SIZE); break; // This calculates the answer exactly
```

CMakeLists.txt
```diff
@@ -31,8 +31,8 @@ endmacro()
 # the final executable name
 set(EXE_NAME babelstream)
 
-# for chrono and some basic CXX features, models can overwrite this if required
-set(CMAKE_CXX_STANDARD 11)
+# for chrono, make_unique, and some basic CXX features, models can overwrite this if required
+set(CMAKE_CXX_STANDARD 14)
```
tomdeakin, line 35: I'm happy to bump to C++14, that sounds a reasonable step. Compiler support seems pretty widespread at this point.

src/main.cpp
```diff
@@ -203,393 +150,226 @@ void run()
...
 }
 
-int parseUInt(const char *str, unsigned int *output)
+void parseArguments(int argc, char *argv[])
 {
-  char *next;
-  *output = strtoul(str, &next, 10);
-  return !strlen(next);
-}
+  auto parseUInt =[](const char *str, size_t *output) {
```
tomdeakin, line 407:
```suggestion
  auto parseUInt = [](const char *str, size_t *output) {
```

src/main.cpp
```diff
@@ -203,393 +150,226 @@ void run()
...
+    char *next;
+    *output = strtoull(str, &next, 10);
+    return !strlen(next);
+  };
+  auto parseInt =[](const char *str, intptr_t *output) {
```
tomdeakin:
```suggestion
  auto parseInt = [](const char *str, intptr_t *output) {
```

src/main.cpp
```diff
@@ -660,18 +496,24 @@ void parseArguments(int argc, char *argv[])
...
       std::cout << "  -n  --numtimes   NUM     Run the test NUM times (NUM >= 2)" << std::endl;
       std::cout << "      --float              Use floats (rather than doubles)" << std::endl;
-      std::cout << "      --triad-only         Only run triad" << std::endl;
-      std::cout << "      --nstream-only       Only run nstream" << std::endl;
+      std::cout << "  -o  --only       NAME    Only run one benchmark (see --print-names)" << std::endl;
+      std::cout << "      --print-names        Prints all available benchmark names" << std::endl;
       std::cout << "      --csv                Output as csv table" << std::endl;
+      std::cout << "      --megabytes          Use MB=10^6 for bandwidth calculation (default)" << std::endl;
       std::cout << "      --mibibytes          Use MiB=2^20 for bandwidth calculation (default MB=10^6)" << std::endl;
+      std::cout << "      --gigibytes          Use GiB=2^30 for bandwidth calculation (default MB=10^6)" << std::endl;
```
tomdeakin:
```suggestion
      std::cout << "      --gibibytes          Use GiB=2^30 for bandwidth calculation (default MB=10^6)" << std::endl;
```

tomdeakin (changes requested):
Thanks for the refactor, @gonzalobg! This is looking great - just the ordering of the execution to invert in two places and then I think it's good to go.

src/main.cpp
gonzalobg: The last commit implements an `--order` option to select the order (classic vs isolated) and sets its default to "classic" which is the order mentioned here.

gonzalobg:
@tomdeakin finished the changes :)

src/main.cpp
```diff
@@ -203,393 +183,246 @@ void run()
...
+
+  switch(order) {
+  // Classic runs each benchmark once in the order specifies in the "bench" array above,
+  // and then repeats num_times:
+  case BenchOrder::Classic: {
+    for (size_t k = 0; k < num_times; k++) {
+      for (size_t i = 0; i < num_benchmarks; ++i) {
+	if (!run_benchmark(bench[i])) continue;
+	run(i);
+      }
```
tomdeakin:
```suggestion
      for (size_t i = 0; i < num_benchmarks; ++i) {
	      if (!run_benchmark(bench[i])) continue;
	      run(i);
      }
```

src/main.cpp
```diff
@@ -632,21 +467,89 @@ void parseArguments(int argc, char *argv[])
...
+      {
+        auto p = std::find_if(bench.begin(), bench.end(), [&](Benchmark const& b) {
+	  return std::string(b.label) == key;
+        });
+        if (p == bench.end()) {
+          std::cerr << "Unknown benchmark name \"" << argv[i] << "\" after --only" << std::endl;
+          std::cerr << "Available benchmarks: All, Classic,";
+          print_labels(std::cerr);
+          std::cerr << std::endl;
+	  std::exit(EXIT_FAILURE);
```
tomdeakin:
```suggestion
          std::cerr << std::endl;
          std::exit(EXIT_FAILURE);
```

src/main.cpp
```diff
@@ -632,21 +467,89 @@ void parseArguments(int argc, char *argv[])
...
+        }
+        selection = p->id;
+      }
+    }
+    else if (!std::string("--order").compare(argv[i]))
+    {
+      if (++i >= argc)
+      {
+	std::cerr << "Expected benchmark order after --order. Options: \"classic\" (default), \"isolated\"."
+		  << std::endl;
```
tomdeakin:
```suggestion
       std::cerr << "Expected benchmark order after --order. Options: \"classic\" (default), \"isolated\"."
		  << std::endl;
```

src/main.cpp
```diff
@@ -632,21 +467,89 @@ void parseArguments(int argc, char *argv[])
...
+      }
+      else if (key == "All")
+      {
+        selection = BenchId::All;
+      }
+      else
+      {
+        auto p = std::find_if(bench.begin(), bench.end(), [&](Benchmark const& b) {
+	  return std::string(b.label) == key;
+        });
```
tomdeakin:
```suggestion
        auto p = std::find_if(bench.begin(), bench.end(), [&](Benchmark const& b) {
          return std::string(b.label) == key;
        });
```

src/main.cpp
```diff
@@ -632,21 +467,89 @@ void parseArguments(int argc, char *argv[])
...
+      exit(EXIT_SUCCESS);
     }
-    else if (!std::string("--nstream-only").compare(argv[i]))
+    else if (!std::string("--only").compare(argv[i]) || !std::string("-o").compare(argv[i]))
     {
-      selection = Benchmark::Nstream;
+      if (++i >= argc)
+      {
+	std::cerr << "Expected benchmark name after --only" << std::endl;
+        exit(EXIT_FAILURE);
```
tomdeakin:
```suggestion
      {
        std::cerr << "Expected benchmark name after --only" << std::endl;
        std::exit(EXIT_FAILURE);
```

src/main.cpp
```diff
@@ -632,21 +467,89 @@ void parseArguments(int argc, char *argv[])
...
       use_float = true;
     }
-    else if (!std::string("--triad-only").compare(argv[i]))
+    else if (!std::string("--print-names").compare(argv[i]))
     {
-      selection = Benchmark::Triad;
+      std::cout << "Available benchmarks: ";
+      print_labels(std::cout);
+      std::cout << std::endl;
+      exit(EXIT_SUCCESS);
```
tomdeakin:
```suggestion
      std::exit(EXIT_SUCCESS);
```

src/main.cpp
```diff
@@ -608,11 +441,13 @@ void parseArguments(int argc, char *argv[])
     else if (!std::string("--arraysize").compare(argv[i]) ||
              !std::string("-s").compare(argv[i]))
     {
-      if (++i >= argc || !parseInt(argv[i], &ARRAY_SIZE) || ARRAY_SIZE <= 0)
+      intptr_t array_size;
+      if (++i >= argc || !parseInt(argv[i], &array_size) || array_size <= 0)
       {
         std::cerr << "Invalid array size." << std::endl;
         exit(EXIT_FAILURE);
```
tomdeakin: `std::`

tomdeakin (changes requested):
Ordering LGMT now - thanks.
Just some weird mixed tab/space issues causes misaligned indenting so we need to run a tabs->spaces conversion to fix that.

src/main.cpp
gonzalobg:
```suggestion
        std::exit(EXIT_FAILURE);
```

gonzalobg:
> Ordering LGMT now - thanks. Just some weird mixed tab/space issues causes misaligned indenting so we need to run a tabs->spaces conversion to fix that.

@tomdeakin fixed.

tomdeakin:
Thanks @gonzalobg !

PR #188, opened 2024-03-14
Add support for 64-bit indices and large arrays

gonzalobg:
This PR adds support for 64-bit indices and large arrays. It is blocked on merging #186 .
Tested the following up to 141 GB of memory:
- cuda
- thrust
- std-data/indices/ranges
- omp
- raja
- tbb
- ocl

The following fail, and I need to minimize and report a bug; this PR does not regress them for small sizes:
- kokkos
- acc 

The following is not tested:
- hip: @tom91136 it would be good to test it and test whether `intptr_t` or `size_t` is better for the indices (or no difference).
- sycl

### DIFF
--- a/src/hip/HIPStream.cpp
+++ b/src/hip/HIPStream.cpp
@@ -25,7 +25,7 @@ void check_error(void)
 __host__ __device__ constexpr size_t ceil_div(size_t a, size_t b) { return (a + b - 1)/b; }
 
 template <class T>
-HIPStream<T>::HIPStream(const int ARRAY_SIZE, const int device_index)
+HIPStream<T>::HIPStream(const intptr_t ARRAY_SIZE, const int device_index)
 {
   // Set device
   int count;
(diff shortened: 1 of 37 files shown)


src/hip/HIPStream.cpp
```diff
@@ -107,9 +107,9 @@ HIPStream<T>::~HIPStream()
 
 
 template <typename T>
-__global__ void init_kernel(T * a, T * b, T * c, T initA, T initB, T initC, int array_size)
+__global__ void init_kernel(T * a, T * b, T * c, T initA, T initB, T initC, intptr_t array_size)
```
gonzalobg: Change this to size_t to make it equal to CUDA for now. Need to test this on AMD GPUs.

gonzalobg:
@tomdeakin @tom91136 This is blocked on #186, but is rebased on it now, and I've fixed the last thing that TomD requested, so should be ready once CI is green.

tomdeakin:
@gonzalobg - think this is good to merge now too, right?

gonzalobg:
Yes, I've rebased this on top of #186, and all tests also pass :)

Issue #191, opened 2024-04-16
Support clang++ for CUDA


tom91136:
Clang++ supports CUDA, we'll need to tweak the CMake models to accommodate this.

PR #193, opened 2024-04-16
Add CUDA clang++ support

tom91136:
clang++ supports compiling CUDA sources directly and the resulting binary works as expected for the CUDA model.
This PR adds a `-DCUDA_CLANG_DRIVER=ON|OFF`  option that defaults to OFF so that nvcc continues to work.
The feature will required CMake >= 3.18: https://cmake.org/cmake/help/latest/release/3.18.html#languages

Addresses #191


tomdeakin:
The CI failed for this on the NVC++ compiler.

PR #195, opened 2024-05-02
Add hipstdpar support to BabelStream

gsitaram:
This PR adds support for offload to AMD GPUs using the `par_unseq` execution policy in C++ standard parallelism algorithms. To trigger the GPU offload of all parallel algorithms, the `--hipstdpar` compilation flag must be provided. For GPU targets other than the current default of `gfx906`, the `--offload-arch=<arch_string>` option must also be provided at compile time.
 
When using ROCm 6.1.0, the compilation commands may look like the following if compiling for an AMD Instinct MI200 series GPU:
 
```
cmake -Bbuild -H. -DMODEL=std-data -DCMAKE_CXX_COMPILER=hipcc -DCLANG_OFFLOAD=gfx90a
cmake --build build
```
 
Remember to set the environment variable to enable address translation and page migration (where applicable) when running `std-data-stream` or `std-indices-stream`:
 
```
export HSA_XNACK=1
```


src/std-indices/model.cmake
```diff
@@ -50,4 +66,13 @@ macro(setup)
         register_definitions(USE_ONEDPL)
         register_link_library(oneDPL)
     endif ()
+    if (CLANG_OFFLOAD)
```
tomdeakin: I'm not totally sold on the name of this definition as CLANG is not specific to AMD, but the `CLANG_FLAGS` set below are definitely specific. Can we rename this please to something appropriate?

tomdeakin (changes requested):

tomdeakin:
It's great to see hipstdpar working, so let's work to get this merged in. Thanks for the contributions.

src/std-indices/model.cmake
gsitaram: Would replacing `CLANG_OFFLOAD` with `AMDGPU_TARGET_OFFLOAD` and `CLANG_FLAGS` with `AMDGPU_TARGET_OFFLOAD_FLAGS` be acceptable?

gsitaram: Hi @tomdeakin, please confirm if the above suggestion would work or recommend something better. I can make the changes accordingly.

gsitaram:
Hi @tomdeakin, @afanfa and I have made the changes requested. Please check and approve if everything looks okay. 
Thanks!

gonzalobg:
Added this PR with some fixes to #202 .

src/std-indices/model.cmake
tomdeakin: LGTM

tomdeakin (approved):

tomdeakin:
Closing as this will be merged as part of #202 and working to merge that.

PR #199, opened 2024-05-30
Cleanup CUDA implementation a bit

gonzalobg:
- Refactor all kernels into a generic "parallel for" algorithm
  - Supports grid-stride and block-stride loops, configurable with model flag
  - Handles devices of different sizes via occupancy APIs
- Refactor memory allocation APIs
- Prints more GPU details, in particular, the theoretical peak BW in GB/s of the current device, using the NVML library (which is part of the CUDA Toolkit and always available)
- Fixes 2 bugs:
  - Prints the "order" used to run the benchmarks (e.g. classic vs isolated)
  - Fixes a division by zero bug in the solution checking

### DIFF
--- a/src/cuda/CUDAStream.cu
+++ b/src/cuda/CUDAStream.cu
@@ -5,18 +5,25 @@
 // source code
 
 #include "CUDAStream.h"
+#include <nvml.h>
 
 #if !defined(UNROLL_FACTOR)
 #define UNROLL_FACTOR 4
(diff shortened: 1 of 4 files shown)


gonzalobg:
This was passing. Seems like this and other PRs are spuriously failing due to some cache issue @tom91136 @tomdeakin

gonzalobg:
Closing for #202

Issue #128, opened 2022-04-04
Update verification check to save memory


tomdeakin:
The OpenMP CPU version allocates twice the memory it needs restricting the maximum problem size.

For offload models, this is OK as the data needs to exists on the device during the run and on the host during verification.

This could be solved by asking the implementations to pass a pointer to host data back to the driver routine for checking, rather than copying into a vector allocated by the driver.
This would allow implementations to do save memory where possible.

tomdeakin:
Refers to this: https://github.com/UoB-HPC/BabelStream/blob/7d570547edf0f0222d022d8cd1d67f768e666738/src/main.cpp#L330C3-L330C3

tomdeakin:
Think this is taken care of by #202

PR #202, opened 2024-06-03
Cleanup CUDA, Reuse Memory, Add Serial Model, Cleaup Std Parallelism

gonzalobg:
# Cleanup CUDA

- Refactor all kernels into a generic "parallel for" algorithm that supports grid-stride and block-stride loops, configurable with model flag.
- Use Occupancy APIs to portably handle devices of all sizes.
- Refactor CUDA memory allocation APIs.
- Prints more GPU details, in particular, the theoretical peak BW in GB/s of the current device, using the NVML library (which is part of the CUDA Toolkit and always available)
- Fixes 2 bugs:
  - Prints the "order" used to run the benchmarks (e.g. classic vs isolated)
  - Fixes a division by zero bug in the solution checking

# Add Serial

By @tom91136 Good thing to have when comparing with other parallel programming models, mostly for syntax.
This also makes us consistent with CloverLeaf, TeaLeaf, and miniBUDE.

# Reuse Memory

This PR puts benchmarks in control of allocating the host
memory used for verifying the results.

This enables benchmarks that use Unified Memory for the device
allocations, to avoid the host-side allocation and just pass
pointers to the device allocation to the benchmark driver.

Closes https://github.com/UoB-HPC/BabelStream/issues/128 .

# Cleanup C++ Standard Parallelism

Merge the 3 implementations into one with different flags for data c++17, data c++23, and indices.
Also annotate workarounds with a `#define WORKAROUND` and print whether the current implementation is not conforming.
Adds support for AdaptiveCpp (CI not added yet; will be done later as part of removing hipSYCL).


src/std/STDStream.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include <ranges>
+#endif
+
+ // OneDPL workaround; TODO: remove this eventually
+#include "dpl_shim.h"
+
+#ifdef INDICES
+// NVHPC workaround: TODO: remove this eventually 
+#if defined(__NVCOMPILER) && defined(_NVHPC_STDPAR_GPU)
+#define WORKAROUND
```
gonzalobg, line 21: Have a pragma message to print that workarounds are enabled.

src/dpl_shim.h
```diff
@@ -29,12 +29,15 @@ T *alloc_raw(size_t size) { return sycl::malloc_shared<T>(size, exe_policy.queue
...
 
+#define WORKAROUND
+
 #else
 
 // auto exe_policy = dpl::execution::seq;
 // auto exe_policy = dpl::execution::par;
 static constexpr auto exe_policy = dpl::execution::par_unseq;
 #define USE_STD_PTR_ALLOC_DEALLOC
+#define WORKAROUND
```
gonzalobg, line 40: pragma message to highlight that there is a workaround

src/sycl2020/SYCLStream2020.cpp
```diff
@@ -1,5 +1,5 @@
 
-// Copyright (c) 2015-23 Tom Deakin, Simon McIntosh-Smith, and Tom Lin
+// Copyright (c) 2015-16 Tom Deakin, Simon McIntosh-Smith,
```
gonzalobg: Undo this change

tomdeakin:
```suggestion
// Copyright (c) 2015-23 Tom Deakin, Simon McIntosh-Smith, and Tom Lin
```

tomdeakin:
The CI is failing during setup - investigating.

PR #208, opened 2024-07-03
Use thrust::transform overload for two ranges where possible

bernhardmgruber:
The use of `zip_iterator`s is only needed beyond 2 input ranges for `thrust::transform`. This PR proposes a corresponding simplification.


bernhardmgruber:
Ping.

PR #209, opened 2024-07-03
Avoid checking for thrust TBB backend twice

bernhardmgruber:
This PR removes a redundant check for the TBB backend in the Thrust version.


bernhardmgruber:
Ping.

tomdeakin:
~This check is looked fro two different Thrust backends. Does `THRUST_DEVICE_SYSTEM_TBB` no longer exist?~
Apologies, GitHub didn't show me the full context. 
This looks good, thanks for the PR!

PR #214, opened 2025-03-30
Add missing <cstdint> and <cstddef> headers.

breyerml:
This PR fixes a bug introduced in #188 due to not adding the necessary `<cstdint>` headers when using `inptr_t`.

Additionally, as far as I understand the code, the current develop branch exhibits UB or, at best, implementation-defined behavior due to the new error checking introduced in #186. If I only enable, for example, the triad kernel, `goldC` is never written and, therefore, is equal to 0 (the value of `startC` in the `Stream.h` header). However, later, the error is calculate as 
```c++
T eC = std::fabs(vC - goldC) / std::fabs(goldC);
```
, which results in a division by zero and sets `eC` to NaN. In the `check` function, `eC` is tested against NaN and reports a failing check, although `vC` is **also** zero, in which case the result should be marked as correct. 
I added a (ugly) hotfix, but I'm currently unable to implement a better fix due to time constraints.

### DIFF
--- a/src/StreamModels.h
+++ b/src/StreamModels.h
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdint>
 #include <memory>
 
 #if defined(CUDA)
--- a/src/acc/ACCStream.h
+++ b/src/acc/ACCStream.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <iostream>
 #include <stdexcept>
 
--- a/src/cuda/CUDAStream.h
+++ b/src/cuda/CUDAStream.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <iostream>
 #include <stdexcept>
 #include <sstream>
--- a/src/hip/HIPStream.h
+++ b/src/hip/HIPStream.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <iostream>
 #include <stdexcept>
 #include <sstream>
--- a/src/kokkos/KokkosStream.hpp
+++ b/src/kokkos/KokkosStream.hpp
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <iostream>
 #include <stdexcept>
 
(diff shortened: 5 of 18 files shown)


breyerml:
Also, if I run:
```
./omp-stream -s 10000000000 -n 2 --only Triad
```
all I get is a SegFault.
However, running it with
```c++
./omp-stream -s 1000000000 -n 2 --only Triad
```
gives to correct result. 

(Note that our machine has enough RAM for the first call.)

breyerml:
Ok. The problem with this SegFault is that due to the `read_arrays` function, the STREAM data must be stored **twice**. Therefore, we actually have `6 * sizeof(T) * array_size` instead of `3 * sizeof(T) * array_size`, which is kind of unintuitive, to say the least.

tomdeakin:
> Ok. The problem with this SegFault is that due to the `read_arrays` function, the STREAM data must be stored **twice**. Therefore, we actually have `6 * sizeof(T) * array_size` instead of `3 * sizeof(T) * array_size`, which is kind of unintuitive, to say the least.

This is a requirement because some models need to own their own data, either intrinsically or for performance reasons. Ideally we'd not need two copies, so would need to migrate all checking routines to each model. See  #128

tomdeakin (commented):
~I'm not sure I understand your comments, as `goldC` etc are initialised no matter what mode is run.~
https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L518
OK, the problem is only with `goldC` because it is initialised to zero. I think we need a nicer solution to ensure that this is checked properly.

tomdeakin:
> ~I'm not sure I understand your comments, as `goldC` etc are initialised no matter what mode is run.~
> 
> https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L518
> 
> 
> OK, the problem is only with `goldC` because it is initialised to zero. I think we need a nicer solution to ensure that this is checked properly.

Also fixed in #202

tomdeakin:
Closing as fixed in #202 and working to merge that to `develop`

PR #217, opened 2026-02-11
Add optional support for GPU page faults in SYCL2020 and OMP backends

ifdu:


tomdeakin (changes requested):
The OpenMP version needs to use the `requires(unified_shared_memory)` directive for correctness according to the standard. Please can you update the change to use this?

I do not believe the SYCL change here is standards compliant to SYCL 2020. Please can you remove this change.

ifdu:
@tomdeakin Thanks for reviewing.

Sure, let me add that directive to the OpenMP version.

The SYCL change is an attempt to make use of [1]:

> In the most capable systems, users do not need to use SYCL USM allocation functions to create shared allocations. The system allocator (malloc/new) may instead be used. Likewise, std::free and delete are used instead of sycl::free.

Would you have a recommendation to make it compliant? 

[1] https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_usm_allocations

tomdeakin:
Good point, thanks for reminding me of that sentence. I don't think SYCL has a way to programmatically check for system capability, so the solution you have is probably as good as we can do for now. Thanks!

ifdu:
@tomdeakin This pull request is now rebased on [UoB-HPC:develop](https://github.com/UoB-HPC/BabelStream/tree/develop) and includes the OpenMP directive which was missing.

src/omp/OMPStream.cpp
```diff
@@ -173,19 +187,23 @@ void OMPStream<T>::triad()
...
-  T *c = this->c;
-  #pragma omp target teams distribute parallel for simd
+  #if !defined(PAGEFAULT)
+    intptr_t array_size = this->array_size;
+    T *a = this->a;
+    T *b = this->b;
+    T *c = this->c;
+    #pragma omp target teams distribute parallel for simd
+  #else
+    #pragma omp target teams distribute parallel for simd is_device_ptr(a, b, c)
```
tomdeakin: I'm not 100% sure whether the `is_device_ptr` clause is required here when you have the `unified_shared_memory` enabled... What was the reason for doing this?

tomdeakin (changes requested):
SYCL changes are fine. Just one query about the OpenMP `is_device_ptr` clause I added to a comment. Thanks for working through this!

src/omp/OMPStream.cpp
ifdu: It is a leftover from early tests but you are correct, it is not needed.

ifdu:
@tomdeakin I pushed a new OpenMP version without `is_device_ptr`, which works fine in my tests.

tomdeakin (approved):

tomdeakin:
Something is causing the CI to fail vs the `develop` branch (https://github.com/UoB-HPC/BabelStream/actions/runs/21986408147/job/63521614186). I will try to investigate.

ifdu:
Does it come from the change or from the CI environment itself?

tomdeakin:
Looks like the environment.

PR #222, opened 2026-06-11
Improved CUDA performance through pipelined reads

rietmann-nv:
This PR improves performance of all CUDA stream examples implemented through pipelined reads, particularly on Blackwell compute GPUs like GB200. For large enough arrays, we see performance of well over 7TB/s, which is much closer to the theoretical bandwidth available. I increased the default array size, to better saturate modern devices.

In this PR, I also fixed the compilation of the thrust version, which didn't get updated when the Stream interface changed.

I also made a small fix for compilation on CUDA 13.2, which can't have `-DDEFAULT` due to a compatibility issue with CCCL. I changed it to `-DBABEL_DEFAULT`.

Very open to feedback, thanks!

### DIFF
--- a/src/cuda/model.cmake
+++ b/src/cuda/model.cmake
@@ -3,11 +3,12 @@ register_flag_optional(CMAKE_CXX_COMPILER
         "Any CXX compiler that is supported by CMake detection, this is used for host compilation"
         "c++")
 
+# "DEFAULT" define causes a compile error in newer cuda CCCL, so we change to BABEL_DEFAULT
 register_flag_optional(MEM "Device memory mode:
-        DEFAULT   - allocate host and device memory pointers.
+        BABEL_DEFAULT   - allocate host and device memory pointers.
         MANAGED   - use CUDA Managed Memory.
         PAGEFAULT - shared memory, only host pointers allocated."
-        "DEFAULT")
+        "BABEL_DEFAULT")
 
 register_flag_optional(STRIDE "Kernel stride: GRID_STRIDE or BLOCK_STRIDE" "GRID_STRIDE")
 
--- a/src/thrust/ThrustStream.cu
+++ b/src/thrust/ThrustStream.cu
@@ -31,13 +31,8 @@ using vector =
 template <class T>
 struct ThrustStream<T>::Impl{
   vector<T> a, b, c;
+  thrust::host_vector<T> h_a, h_b, h_c;
 };
-
-template <class T>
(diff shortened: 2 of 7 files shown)


src/thrust/ThrustStream.cu
```diff
@@ -32,6 +33,12 @@ struct ThrustStream<T>::Impl{
   vector<T> a, b, c;
 };
 
+template <class T>
+struct ThrustStream<T>::h_Impl{
+  thrust::host_vector<T> a, b, c;
+};
```
bernhardmgruber: Suggestion: let's just add the new vectors to `Impl`:

```suggestion
template <class T>
struct ThrustStream<T>::Impl{
  vector<T> a, b, c;
#if !(defined(PAGEFAULT) || defined(MANAGED))
  // we need separate host allocations to hold the data for get_arrays()
  thrust::host_vector<T> host_a, host_b, host_c;
#endif
};
```

src/thrust/ThrustStream.h
```diff
@@ -20,11 +20,13 @@ class ThrustStream : public Stream<T>
 {
   protected:
     struct Impl;
+    struct h_Impl;
     std::unique_ptr<Impl> impl; // avoid thrust vectors leaking into non-CUDA translation units
+    std::unique_ptr<h_Impl> h_impl; // If UVM is disabled, host arrays for verification purposes
```
bernhardmgruber: Not needed, if the host vectors are moved into `Impl`

src/cuda/model.cmake
```diff
@@ -3,11 +3,12 @@ register_flag_optional(CMAKE_CXX_COMPILER
...
         "c++")
 
+# "DEFAULT" define causes a compile error in newer cuda CCCL, so we change to BABEL_DEFAULT
 register_flag_optional(MEM "Device memory mode:
-        DEFAULT   - allocate host and device memory pointers.
+        BABEL_DEFAULT   - allocate host and device memory pointers.
         MANAGED   - use CUDA Managed Memory.
         PAGEFAULT - shared memory, only host pointers allocated."
-        "DEFAULT")
+        "BABEL_DEFAULT")
```
bernhardmgruber, line 11: Important: That's a bug in CCCL, if we are senstive to that macro. Please file an issue. Maybe @miscco can have a look at that.

bernhardmgruber: That's resolved now in upstream CCCL: https://github.com/NVIDIA/cccl/issues/9406

PR #225, opened 2026-08-11
Centralize host allocation defaults across CPU backends

### LINKED ISSUE #201: Refactor malloc allocation
Currently, some of the benchmarks allocate memory that's accessed from the kernels using:
- `malloc`
- `aligned_alloc(2MiB, ...)`
- C++ `new`
- vendor specific apis (cudaMallocManaged, etc.)

How the memory is allocated does impact benchmark performance a bit. 

I think we should extract these into their own file, to ensure all benchmarks may pick the exact same defaults.

mohitt31:
## What this does

As mentioned by @gonzalobg in #201, allocations were scattered across `malloc`, `aligned_alloc(2MiB)`, and `new`. This PR consolidates the host allocation defaults into one place so all models pick up the same behavior.

* Added `src/Alloc.h` containing `ALIGNMENT`, `alloc_raw`, and `dealloc_raw`.
* `omp`, `raja`, `serial`, and `tbb` now use these shared helpers.
* The function names match what `dpl_shim.h` was already using, so the `std` model didn't require any code changes.
* `ALIGNMENT` is still wrapped in `#ifndef`, so overriding it via `-DALIGNMENT=...` works exactly like before.

## Bug fix: macOS `aligned_alloc` crash

`aligned_alloc` requires the requested size to be a multiple of the alignment. `glibc` is forgiving and lets unrounded sizes pass, but other libcs (like on macOS) return `NULL` instead. Since there was no null check, running `./serial-stream -s 1000` or `-s 1048577` on macOS caused a segfault (exit 139). It went unnoticed with the default size (`33554432`) because that happens to be a perfect multiple of 2MB.

* The size is now rounded up centrally in one place, fixing this for all models.
* A failed allocation now throws a `std::runtime_error` instead of returning null and crashing (this also matches how the HIP and CUDA models currently handle memory errors).

## Testing

* Environment: macOS 26.5, Apple clang, arm64.
* Built `serial` and `omp`.
* Ran 12 tests: 3 sizes (`1000`, `1048577`, `33554432`) × `double` and `float`. All passed with zero validation failures. (For context, the `1000` and `1048577` runs segfault on `main`).
* Performance: Triad default size was ~95,590 MB/s on `main`, and ~96,212 MB/s on this branch. No real difference, just normal noise.
* Untested locally: `raja`, `tbb`, and `std` (I don't have this toolchain set up on my Mac). The changes for `raja` and `tbb` are identical direct call swaps. (`std` doesn't build on my machine anyway since Apple clang lacks `<execution>`).

## Note on `sycl2020`

The `sycl2020` model has the exact same duplicated `ALIGNMENT` logic, but since PR #226 is actively rewriting that file, I intentionally left it out of this PR to avoid merge conflicts. I will do a quick follow-up PR for it once #226 is merged.

### DIFF
--- a/src/Alloc.h
+++ b/src/Alloc.h
@@ -1,0 +1,39 @@
+// Copyright (c) 2015-16 Tom Deakin, Simon McIntosh-Smith,
+// University of Bristol HPC
+//
+// For full license terms please see the LICENSE file distributed with this
+// source code
+
+#pragma once
+
+#include <cstddef>
+#include <cstdlib>
+#include <stdexcept>
+#include <string>
+
+// Host allocation defaults, shared by every model that allocates its arrays on
+// the host, so they all use the same alignment.
+
+#ifndef ALIGNMENT
+#define ALIGNMENT (2*1024*1024) // 2MB
+#endif
+
+template <class T>
+T *alloc_raw(size_t size)
+{
+  // aligned_alloc requires the size to be a multiple of the alignment, so round
+  // it up. glibc lets the unrounded size through, but other libcs return a null
+  // pointer, which used to show up as a crash for array sizes that are not a
+  // multiple of ALIGNMENT.
+  size_t bytes = sizeof(T) * size;
+  size_t rounded = ((bytes + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
+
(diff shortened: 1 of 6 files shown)


mohitt31:
ping @tomdeakin to review this when you get a chance

PR #226, opened 2026-09-03
[SYCL2020] Support device USM

EwanC:
Add device USM support to the SYCL2020 model for SYCL devices that don't support shared USM.

This patch also cleans up the build infrastructure for the SYCL2020 model which currently fails to build. In particular:

* Makes consistent in doc and cmake that `-DMODEL=sycl2020` should be passed rather than `sycl2020-usm` or `sycl2020-acc`.
* Removes buffer/accessors from model, as this code is broken in several ways that means it won't compile. For example, using default buffer constructor and not being passed to the reduction correctly. USM is the preferred memory model in SYCL2020 so I think only using it is fine as buffer/accessors are used in the other SYCL 1.2 model.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,12 +184,6 @@ else ()
     message(STATUS "Selected model  :  ${MODEL}")
 endif ()
 
-if (MODEL STREQUAL "sycl2020")
-    message(FATAL_ERROR "
-    Model sycl2020 has been renamed to sycl2020-acc, and a new sycl2020-usm model is now available.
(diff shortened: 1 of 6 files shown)


EwanC:
ping @tomdeakin to review this

EwanC:
@tom91136 @miloszwasacz You are the only two members I can see in the UoB-HPC organization people list (https://github.com/orgs/UoB-HPC/people), would you be able to review this PR and also the Vulkan one I have opened since?

